PR #56, opened 2024-04-13
Multilingual LLava bench 

gagan3012:
This is the task to run multilingual llava bench from [MBZUAI](https://huggingface.co/datasets/MBZUAI/multilingual-llava-bench-in-the-wild)


Luodian:
Thanks for this PR, it's clean since it's only adding files in `tasks` folder.

May I ask that if you could also give a screenshot to see a standard model (e.g. llava-v1.5-7b) results on multilingual-llava-bench?

Luodian (approved):

PR #61, opened 2024-04-22
Fix typo in Qwen-VL that was causing "reference before assignment"

tupini07:
Hi! This PR fixes an issue in the Qwen-VL `generate_until` method that was caused by a typo (`contexts[i]` was being referenced as `context[i]`; singular). I also added an `if` statement that converts `contexts` to a list so we can use _item assignment_ with the indexing operator. Finally, made a minor refactor to make the construction of the `query` more readable.

### DIFF
--- a/lmms_eval/models/qwen_vl.py
+++ b/lmms_eval/models/qwen_vl.py
@@ -239,13 +239,14 @@ class Qwen_VL(lmms):
             # Similar to llava, is visual paths has len 0
             # Then nothing will be executed
             query = []
-            for visual_path, context in zip(visual_paths, contexts):
-                query.append({"image": visual_path})
-                query.append({"text": context})
-
             if len(visual_paths) == 0:
                 for context in contexts:
                     query.append({"text": context})
+            else: 
+                for visual_path, context in zip(visual_paths, contexts):
+                    query.append({"image": visual_path})
+                    query.append({"text": context})
+
 
             questions = self.tokenizer.from_list_format(query)
             input_ids = self.tokenizer(questions, return_tensors="pt", padding="longest")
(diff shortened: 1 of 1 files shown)


lmms_eval/models/qwen_vl.py
```diff
@@ -228,21 +228,25 @@ def _collate(x):
...
                 elif not isinstance(until, list):
                     raise ValueError(f"Expected `gen_kwargs['until']` to be of type Union[str,list] but got {type(until)}")
+
+            if isinstance(contexts, tuple):
+                contexts = list(contexts)
+
             for i in range(len(contexts)):
                 if "<image>" in contexts[i]:
-                    context[i] = contexts[i].replace("<image>", "")
-            questions = [self.prompt.format(visual_path, context) for visual_path, context in zip(visual_paths, contexts)]
```
tupini07, line 234: I removed this line since `questions` is being assigned below in the `questions = self.tokenizer.from_list_format(query)` line.

lmms_eval/models/qwen_vl.py
```diff
@@ -228,21 +228,25 @@ def _collate(x):
                     until = [until]
                 elif not isinstance(until, list):
                     raise ValueError(f"Expected `gen_kwargs['until']` to be of type Union[str,list] but got {type(until)}")
+
+            if isinstance(contexts, tuple):
+                contexts = list(contexts)
```
tupini07, line 233: not sure if this is the best way to handle this, but considering we use item assignment I thought this was ok. Let me know if you have any better ideas.

kcz358:
Hi @tupini07 , thank you for your contribution.

I think we also met the same issue when running inference with qwen. We solved it in a similar way but I think yours are better. @Luodian

kcz358 (approved):

PR #93, opened 2024-05-27
Add m3exam

Jiawei-Guo:
Add m3exam

### DIFF
--- a/lmms_eval/tasks/m3exam/utils.py
+++ b/lmms_eval/tasks/m3exam/utils.py
@@ -1,0 +1,895 @@
+from collections import defaultdict
+import re
+import ast
+import random
+import numpy as np
+import os
+import json
+import logging
+# from PIL import Image
+import io
+from lmms_eval.tasks._task_utils.file_utils import generate_submission_file
+from datasets import Image, load_dataset
+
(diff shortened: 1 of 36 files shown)


lmms_eval/tasks/m3exam/utils.py
```diff
@@ -0,0 +1,895 @@
...
+    if fill_answer:
+        prompt += str(question['answer_text'])
+    
+    return prompt
+
+def construct_prompt(doc):
+# (lang, method, setting, model, test_question, dev_question):
+    lang = doc["language"]
+    method = "default"
+    setting = "zero-shot"
```
kcz358, line 43: Seems like the setting has been fixed into zero-shot. But I see that you write a lot of if-else statement evaluating the setting is `zero-shot` or not, is it true that there will be more settings in the future?

kcz358 (commented):
Thank you for your contribution!

I think there are too many duplicated utils file. Is it possible to move different mmmu into the same folder and use a single utils? Or at least remove the repeat code. There is no need for creating different folders for mmmu_English, mmmu_French etc. Can you try to put it under the same folder and maybe a yaml file that allow user to run all the task using a group

PR #166, opened 2024-07-26
Fix: None value handler for seedbench-2 evaluation code

skyil7:
Since there're only `model_specific_prompt_kwargs` for `llava` and `gpt4v` in `seedbench2.yaml`, attempting to evaluate other models resulted in a `TypeError: 'NoneType' object is not subscriptable` error due to the absence of this argument.

So I simply added lines of code to handle None value in `seed_doc_to_text()`.


skyil7:
Close due to misunderstanding about the code. I will re-open it soon as I fix it correctly.

PR #174, opened 2024-07-30
Fixed some issues in InternVL family and ScienceQA task.

skyil7:
I found some issues during I evaluate InternVL and InternVL2 on ScienceQA task.

1. Errors in `generate_until()`
    - When the input w/o any visual comes in, `generate_until` returns `IndexError` because they assume there must be at least one image in the input. So I fixed it.
2. Result processing code in SQA
    - Missing `strip().lower()`


kcz358 (approved):
LGTM

PR #182, opened 2024-08-07
[New Updates] LLaVA OneVision Release; MVBench, InternVL2, IXC2.5 Interleave-Bench integration.

Luodian:
We are excited to announce the release of the latest updates for our project, featuring the integration of LLaVA OneVision and new benchmarking tools: MVBench, InternVL2, and IXC2.5 Interleave-Bench.

# Key Features
- LLaVA OneVision Release
  - vanilla inference model
  - sglang srt model
- InternVL2 Integration
- IXC2.5 Interleave-Bench Integration
- MVBench Integration


lmms_eval/models/__init__.py
kcz358, line 1: I think some of the models were deleted when merging. Probably needs to bring back here. Maybe it is better to resolve the conflict in the init file of model manually.

kcz358 (commented):
I think the only issue might be the init for the model. Are these deleted models supposed to remove?

kcz358 (approved):

linkangheng:
When evaluating MVBench, the generate_until function in lmms_eval/models/llava_vid.py cannot handle the situation that the folder is full of frames, such as mvbench/tvqa , and a function is needed to deal with it.

Issue #243, opened 2024-09-11
[BUG] Commit #0d02bad may break `srt_api` by mis-use of asyncio


Maxwell-Lyu:
### When you open an issue, please be sure to include the following

- [x] A descriptive title: [xxx] XXXX
- [x] A detailed description

# Where's the problem

At `generate_until` in `srt_api.py`, at [here](https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/8f9f18edaca0134d44ba9a544b87076539e86e0d/lmms_eval/models/srt_api.py#L217)

The author @kcz358 uses `asyncio.as_completed` in an order to send multiple requests to the server at the same time for faster evaluation, committed as 0d02bad.  

# What's happening
`asyncio.as_completed` is non-blocking, which means this iterator yield whatever request that has a response, regardless of their order of creation.  
So, the order of responses in `res` may not be the order of the `requests`. In fact I should use "probably", because that probability is well above 50% whenever the CPU has more than 4 threads resulting in a `num_processes` of more than 2.
This make `srt_api` more like a rolling dice rather a valid evaluation, giving my colleage a evaluation score of non-sense.

# How to fix
I'm no where near a professional python programmer, but I suggest using `asyncio.gather` instead, which is blocking thus preserves the order.  
Or, if I mis-understood your code? Feel free to correct me and please forgive me if so.

Looking forward for your reply ;)

kcz358:
Yeah I think you are correct. I will try to fix this.

kcz358:
Hi @Maxwell-Lyu , I have created a PR in #244 to fix this issue. The fix is quite simple but there seems somehow some issues in current sglang that might causing the evaluation result to be inconsistent. I have tested that our result can be reproduced by the commit hash `2f1d92834f41df42e266ed6d7036b4add906d21f`. I believe some changes after we PR have changed the behaviour of the model. I will try to check and fix it with the sgl team.

Maxwell-Lyu:
Thank you for your enthusiastic contribution to this project. Have a great day!

PR #244, opened 2024-09-12
[Fix] Fix async append result in different order issue

kcz358:

- [ ] A detailed description

---
As stated in #243 , there is an error in current asyncio srt model. This PR tries to fix this and reproduce the result

### DIFF
--- a/lmms_eval/models/srt_api.py
+++ b/lmms_eval/models/srt_api.py
@@ -40,6 +40,7 @@ class SRT_API(lmms):
         max_frames_num: int = 32,
         timeout: int = 60,
         chat_template: str = "chatml-llava",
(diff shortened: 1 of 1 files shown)


kcz358:
The result can be reproduced now by srt model. The result are mostly aligned with the paper result.

Noted that current sglang main seems to have some issue and can not reproduce the result. I have checked that you are be able to reproduce the result by

```bash
cd sglang
git checkout 2f1d92834f41df42e266ed6d7036b4add906d21f
```

My current result using srt api model on llava-ov-0.5B
## AI2D
image

## MME
image

When using current main branch sglang, I can only get 216/1066 which is relatively low. I will try to check what changes in sglang caused this issue.

Luodian (approved):

PR #249, opened 2024-09-13
[Minor] Update utils.py

YangYangGirl:
correct video path in videomme_dic_to_text+subtitle

### DIFF
--- a/lmms_eval/tasks/videomme/utils.py
+++ b/lmms_eval/tasks/videomme/utils.py
@@ -213,7 +213,8 @@ def videomme_doc_to_text(doc, lmms_eval_specific_kwargs=None):
 
 def videomme_doc_to_text_subtitle(doc, lmms_eval_specific_kwargs=None):
     cache_dir = os.path.join(base_cache_dir, cache_name)
-    video_path = "data/" + doc["videoID"] + ".mp4"
+    video_path = doc["videoID"] + ".mp4"
+    video_path = os.path.join(cache_dir, "data", video_path)
     subtitle_path = os.path.join(cache_dir, "subtitle", doc["videoID"] + ".srt")
     video_path = os.path.join(cache_dir, video_path)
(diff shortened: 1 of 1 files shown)


kcz358:
Hi @YangYangGirl , may I ask why would this additional hardcoded data path need to be added in the doc to visual function? I think current main branch should works well. If that is your default path in your own settings, maybe this change should be add into your own forked repo? If that is the case, I will close the PR.

YangYangGirl:
The original video path in the main repo does not exist, resulting in total_frame = 0 and subtitle loading as none. It will run smoothly but result in inaccurate results.
image

kcz358:
I think this problem has been fixed in commit [Fix videomme evaluation](https://github.com/EvolvingLMMs-Lab/lmms-eval/commit/3726b4f4714b5825bef9028828e985bd878a405d)? Current main branch seems already contains the path to data folder

https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/bfff5abcfbafa28ec5ae9a30418b9d5985606291/lmms_eval/tasks/videomme/utils.py#L176-L188

Seems like a duplicate fix. If this is the case, I will close this PR.

YangYangGirl:
Thanks for the kind reply. But I checked again, in https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/bfff5abcfbafa28ec5ae9a30418b9d5985606291/lmms_eval/tasks/videomme/utils.py#L21, the video path is still wrong.

kcz358:
Yeah @YangYangGirl , I see your point. You are correct. I think this is some version problem as we change the logic and default folder for unzipping files. I revise a little bit to make it not absolute path. I think current version should be correct?

kcz358:
Have done a quick check here and subtitles should have loaded correctly.

image

merging this PR now and thank you for spotting the issue.

kcz358 (approved):

PR #259, opened 2024-09-16
[Feat] add regression test and change saving logic related to `output_path`

Luodian:
This pull request adds new output_path saving logic and a regression test. The new logic ensures that the model's results are saved to the correct output path. Additionally, a regression test has been added to verify that the model's performance is not degraded after each pull request.


kcz358:
Should we merge this? Or we should rewrite the regression test into a unittest format?

PR #260, opened 2024-09-16
[Feat] Add support for llava_hf video, better loading logic for llava_hf ckpt

kcz358:
This PR update the `llava_hf` to enable the evaluation of the `llava_hf` on new series of llava model such as llava-onevision and llava-next(stronger). This PR also enable the video evaluation using llava_hf.

Noted that the video evaluation is only supported using llava onevision hf and would possibly failed if you other version of llava. Since the newest transformers version has not released you have to do

```bash
pip install git+https://github.com/huggingface/transformers.git
```
to install the transformers version from source if you want to use llava onevision hf.

However, after experiment, I think the performance still has some significance difference compare to the original llava. Thus, you are not recommended to use this model to get results of original llava or llava-onevision. This model is only recommended to those that wish to have a quick baseline or have finetuned their model using llava-hf

### DIFF
--- a/lmms_eval/models/llava_hf.py
+++ b/lmms_eval/models/llava_hf.py
@@ -1,11 +1,15 @@
 import warnings
 from typing import List, Optional, Tuple, Union
 
+import numpy as np
+import PIL
 import torch
 from accelerate import Accelerator, DistributedType
 from accelerate.state import AcceleratorState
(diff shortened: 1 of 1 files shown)


kcz358:
Video result aligned for 0.5-ov

image

Luodian (approved):

PR #273, opened 2024-09-22
[Bug] Fix the checking error in gpt4v generation

pufanyi:
API errors should return an empty string instead of the answer from the previous query.


pufanyi:
Sorry that this PR was opened to the wrong destination branch. I will close it.

PR #278, opened 2024-09-24
[Fix] Model name None in Task manager, mix eval model specific kwargs

kcz358:

- [ ] A detailed description

---

We forgot to pass in the model name in the task manager, this could cause the specific kwargs to be useless. This PR is to fix this error


Luodian (approved):
Good fix, thanks!

PR #306, opened 2024-10-08
[Fix] Fix cache_dir issue where MVBench cannot be found

yinanhe:
[Fix] Fix cache_dir issue where MVBench cannot be found
       -  Added Line 1004-1012 in `lmms_eval/api/task.py`. If the downloaded file does not need to be unzipped (unzip or tar), it will directly create the symbolic link the cache path of huggingface to the cache_dir, and the link will be updated after each pull. This operation also applies to dataset-repo that does not need to be unzipped.


yinanhe:
- [feat] New variable in dataset kwargs: `create_link`
  If `create_link` exists in `dataset_kwargs`, a symbolic link will be created to the data repository in cache_dir.
- [fix] Modify the `print` function in the previous code to use `logger`

Luodian (approved):

PR #323, opened 2024-10-15
Add openhermes task

pbcong:


kcz358:
Hi, I have rebase the `feat-dev` and rebase to main so that we can keep up a linear history. Can you do the same to this branch so that only the new commits `add openhermes task` and `formatting` would be squash and merge into the `feat-dev` branch?

If you don't mind I force push on your branch, I can also help you do that

kcz358:
Hi @pbcong , I rebase and keep only these two commits. Can you check current file changes are correct? If everything is correct, I will merge this PR.

pbcong:
Yes, I think the changes are all correct, thank you.

PR #325, opened 2024-10-15
Add AI2D evaluation without masks

zhijian-liu:
This PR adds AI2D evaluation without masks (following VLMEvalKit).


kcz358 (approved):
LGTM, may I ask how this ai2d without mask dataset is different from our dataset version?

zhijian-liu:
@kcz358, the current dataset masks the words in the images and replaces them with options. However, some questions refer specifically to the labels in the images, which can result in lower accuracy when using the masked version. VLMEvalKit includes an option to evaluate AI2D without masking these words, which might be more suitable in such cases.

PR #327, opened 2024-10-16
Update evaluator.py to load datasets first before loading models

LooperXX:
Downloading datasets before loading models will help prioritise data preparation without using GPUs.
- just adjusted the order of the code in evaluator.py
- it has been tested for most of the datasets for single-image and text-only scenarios and no anomalies were found


Luodian (approved):

Luodian:
Hi thanks for this PR, the change is reasonable~ However, it's better to use `pre-commit` to do lint and isort to make all code follow same style.

Issue #350, opened 2024-10-25
Inconsistent evaluation result on Vinoground


Richar-Du:
Thanks for the quick commitment of Vinoground! However, when I use the following command to evaluate LLaVA-OV-7B, the result is lower than the original paper's.
Evaluation command:
```
# video tasks
accelerate launch --main_process_port=29501 --num_processes=8 \
-m lmms_eval \
--model llava_onevision \
--model_args pretrained=/lmms-lab/llava-onevision-qwen2-7b-ov,conv_template=qwen_1_5,model_name=llava_qwen \
--tasks vinoground \
--batch_size 1 \
--log_samples \
--log_samples_suffix llava_onevision \
--output_path ./logs/
```
Evaluation result:
```
action: text: 35.80%, video: 17.51%, group: 8.95%
all: text: 41.20%, video: 18.60%, group: 9.20%
viewpoint: text: 67.47%, video: 25.30%, group: 19.28%
cyclical: text: 32.43%, video: 16.22%, group: 8.11%
spatial: text: 26.21%, video: 14.56%, group: 3.88%
interaction: text: 34.25%, video: 15.07%, group: 4.11%
contextual: text: 34.92%, video: 17.46%, group: 7.94%
object: text: 36.25%, video: 16.88%, group: 4.38%

|  Tasks   |Version|Filter|n-shot|     Metric     |   |             Value             |   |Stderr||----------|-------|------|-----:|----------------|---|-------------------------------|---|------|
|vinoground|Yaml   |none  |     0|vinoground_score|↑  |(41.199999999999996, 18.6, 9.2)|±  |   N/A|
```
<img width="547" alt="image" src="https://github.com/user-attachments/assets/4cce07a9-7338-46a0-ba8d-e67a1d3c8794">

HanSolo9682:
Hi! Thank you for your feedback! We have identified the problem and opened a pull request #354. Now we can correctly reproduce the results.
<img width="995" alt="Screenshot 2024-10-25 at 15 56 13" src="https://github.com/user-attachments/assets/c2843b1b-bb39-42d1-ae4d-bfaceb8017f3">

Richar-Du:
Thanks for your quick feedback!

PR #354, opened 2024-10-25
Update Vinoground to make evaluation consistent with paper

HanSolo9682:
As mentioned by Issue #350, the results of video score and group score on many models are inconsistent with the original paper. This is because in the lmms-eval code we initially provided, we did not use the shuffled questions like we did in the paper. We update the code to correctly reflect that and we are now able to reproduce results on llava-ov-qwen2-7b.

Screenshot 2024-10-25 at 15 56 13


kcz358 (approved):

kcz358:
Thank you for fixing the bugs in the issue! Merging this PR

PR #376, opened 2024-10-29
Add retry for gpt api call and improve air_bench aggregation function

pbcong:
- Issue: Too Many Requests errors in API calls; so I added retry for each call

- Improve air_bench_foundation aggregation to return result for each tasks rather than the total average

### DIFF
--- a/lmms_eval/tasks/clotho_aqa/utils.py
+++ b/lmms_eval/tasks/clotho_aqa/utils.py
@@ -49,9 +49,13 @@ if API_TYPE == "openai":
         "Authorization": f"Bearer {API_KEY}",
         "Content-Type": "application/json",
     }
+
 elif API_TYPE == "azure":
-    API_URL = os.getenv("AZURE_ENDPOINT", "https://api.cognitive.microsoft.com/sts/v1.0/issueToken")
-    API_KEY = os.getenv("AZURE_API_KEY", "YOUR_API_KEY")
+    API_KEY = os.getenv("AZURE_API_KEY", None)
+    deployment_name = os.getenv("AZURE_DEPLOYMENT_NAME", None)
+    resource_name = os.getenv("AZURE_RESOURCE_NAME", None)
+    api_version = os.getenv("AZURE_API_VERSION", None)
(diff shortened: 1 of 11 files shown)


lmms_eval/tasks/clotho_aqa/utils.py
```diff
@@ -49,9 +49,13 @@ def clotho_aqa_doc_to_text(doc, lmms_eval_specific_kwargs):
...
     }
+
 elif API_TYPE == "azure":
-    API_URL = os.getenv("AZURE_ENDPOINT", "https://api.cognitive.microsoft.com/sts/v1.0/issueToken")
-    API_KEY = os.getenv("AZURE_API_KEY", "YOUR_API_KEY")
+    API_KEY = os.getenv("AZURE_API_KEY", None)
+    deployment_name = os.getenv("AZURE_DEPLOYMENT_NAME", None)
+    resource_name = os.getenv("AZURE_RESOURCE_NAME", None)
+    api_version = os.getenv("AZURE_API_VERSION", None)
+    API_URL = f"https://{resource_name}.openai.azure.com/openai/deployments/{deployment_name}/chat/completions?api-version={api_version}"
```
kcz358: A small question on how we format this azure api url. Why don't we simple export the formatted to `API_URL` and then use the original format. This keeps your env clean and less export operation needed.

kcz358 (commented):
Just a small question for how you format the AZURE API URL. Can we just simply export the formatted URL in one go and use the original format. I have seen many different URL format for azure api format so I feel it is important to keep the same

pbcong:
Hi! Thank you for pointing that out. I’ve made the changes to the code.

kcz358 (commented):
Hi, thank you for the changes. You can set the default endpoint to `"https://api.cognitive.microsoft.com/sts/v1.0/issueToken"` if no endpoint is provided. Just like what we did in llava_wilder and many other tasks

https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/df3cb0f0dc42b865e429467b671917575cefa543/lmms_eval/tasks/llava_wilder/utils.py#L42-L48

pbcong:
I updated, thank you!

kcz358 (approved):

PR #381, opened 2024-10-30
Add os import to mathverse_evals.py

### LINKED ISSUE #380: Using Mathverse dataset fails due to import error
`lmms_eval/tasks/mathverse/mathverse_evals.py` is missing the `os` import, which makes it fail when using this dataset.

spacecraft1013:
Fixes #380

### DIFF
--- a/lmms_eval/tasks/mathverse/mathverse_evals.py
+++ b/lmms_eval/tasks/mathverse/mathverse_evals.py
@@ -1,5 +1,5 @@
-import time
 import os
+import time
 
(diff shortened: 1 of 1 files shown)


kcz358:
Hello, I think you need to place the os in the correct position to pass the lint test. You can use the `pre-commit` by following the instructions in the PR doc to do that. Thank you!

spacecraft1013:
Fixed, thanks!

kcz358 (approved):

PR #404, opened 2024-11-09
[Tiny Fix] fix dataset_kwargs in lmms_eval/api/task.py

Li-Qingyun:
It seems that `dataset_kwargs is not None` was lacked.


Li-Qingyun:
The error is like
```
TypeError: argument of type 'NoneType' is not iterable
```

Luodian (approved):

PR #451, opened 2024-12-10
add task MMVet-v2

frankRenlf:
MM-Vet v2 is a extension of MMVet, which includes a new VL capability called "image-text sequence understanding", evaluating models' ability to process VL sequences.
paper: https://arxiv.org/abs/2408.00765
github: https://github.com/yuweihao/MM-Vet/tree/main/v2

Since the sample may have multiple images, it is provided mmvetv2_group_img.yaml for those models that cannot handle multiple image inputs.

### DIFF
--- a/lmms_eval/tasks/mmvetv2/mmvetv2.yaml
+++ b/lmms_eval/tasks/mmvetv2/mmvetv2.yaml
@@ -1,0 +1,27 @@
+dataset_path: lmms-lab/MMVet-v2
+dataset_kwargs:
+  token: True
+task: "mmvetv2"
+test_split: test
+output_type: generate_until
+doc_to_visual: !function utils.mmvet_doc_to_visual
+doc_to_text: !function utils.doc_to_text # Such that {{question}} will be replaced by doc["question"]
+doc_to_target: "{{answer}}"
+generation_kwargs:
+  max_new_tokens: 1024
+  temperature: 0.2
+  top_p: 1.0
+  num_beams: 1
+  do_sample: false
+process_results: !function utils.mmvet_process_results # apply gpt eval here
+metric_list:
+  - metric: gpt_eval_score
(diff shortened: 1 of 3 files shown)


pufanyi:
Hi! Thank you so much for your contribution!!!!

It looks good to me! However, is it possible to add a simple comment before `mmvetv2_group_img.yaml`, something like (if I understood correctly):

```yaml
# This task differs from mmvet2 in that it combines multiple images into one, to accommodate models that cannot accept multiple images as input.
```

Thanks!!!

frankRenlf:
> Hi! Thank you so much for your contribution!!!!
> 
> It looks good to me! However, is it possible to add a simple comment before `mmvetv2_group_img.yaml`, something like (if I understood correctly):
> 
> ```yaml
> # This task differs from mmvet2 in that it combines multiple images into one, to accommodate models that cannot accept multiple images as input.
> ```
> 
> Thanks!!!

Ok, I'll add this comment later

pufanyi:
Thanks!!!

Luodian:
image

Hi I think the evaluation still has some errors?

frankRenlf:
> image
> Hi I think the evaluation still has some errors?

It seems that len of image_list is zero, the images should be stored in doc follow with image name as key. 
This works fine for me.
image

pufanyi:
It seems that it is because there are 2 `doc_to_visual` functions in `utils.py`:

https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/d900496af837dcf5afe3d9efea471d624fd20588/lmms_eval/tasks/mmvetv2/utils.py#L165

https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/d900496af837dcf5afe3d9efea471d624fd20588/lmms_eval/tasks/mmvetv2/utils.py#L297

I have fixed it with some other errors in #457

PR #498, opened 2025-01-13
Fix llama_vision chat_template and decode

coding-famer:
fix adding bos twice and remove eos in decoded results


pufanyi (approved):
Hello! Thank you so much for your contribution!

PR #502, opened 2025-01-16
[FIX] Minor errors in `gemini_api.py` and `internvl2.py`.

skyil7:
This PR fixes two minor errors in models.
## Initialize `self.response_persistent_file` in `gemini_api.py`
- When we use `GeminiAPI` with `continual_mode=False`, It causes ValueError since `self.response_persistent_file` is not declared.
https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/8821357aa338a17b94102118cbb5c8b94f594ae0/lmms_eval/models/gemini_api.py#L58-L64
- So I initialized `self.response_persistent_file` with an empty string (`""`) to prevent error.

### Wrong variable name in `internvl2.py`
- In `internvl2.py`, `self.model.chat()` in line 316 requires `num_patches_list` variable.
- But in line 315, we declare `num_patch_list` instead of `num_patches_list`.
https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/8821357aa338a17b94102118cbb5c8b94f594ae0/lmms_eval/models/internvl2.py#L313-L316
- This seems a typo error, so I fixed it.

BTW, Happy New Year!

### DIFF
--- a/lmms_eval/models/gemini_api.py
+++ b/lmms_eval/models/gemini_api.py
@@ -52,6 +52,7 @@ class GeminiAPI(lmms):
         self.timeout = timeout
         self.model = genai.GenerativeModel(model_version)
         self.continual_mode = continual_mode
+        self.response_persistent_file = ""
         self.interleave = interleave
         # if self.continual_mode and response_persistent_folder is None:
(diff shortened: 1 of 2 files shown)


pufanyi (approved):
Hiiii!!! Thank you so much for your contribution!!! Happy New Year!!!

PR #515, opened 2025-02-03
Add covost2 zh en

pbcong:
covost 2 zh en
  - covost 2 zh en test
  - covost 2 zh en dev

### DIFF
--- a/lmms_eval/tasks/covost2_zh_en/covost2_zh_en_dev.yaml
+++ b/lmms_eval/tasks/covost2_zh_en/covost2_zh_en_dev.yaml
@@ -1,0 +1,3 @@
+task: "covost2_zh_en_dev"
+test_split: dev
+include: _default_template_yaml
--- a/lmms_eval/tasks/covost2_zh_en/covost2_zh_en_test.yaml
+++ b/lmms_eval/tasks/covost2_zh_en/covost2_zh_en_test.yaml
@@ -1,0 +1,3 @@
+task: "covost2_zh_en_test"
+test_split: test
+include: _default_template_yaml
--- a/lmms_eval/tasks/covost2_zh_en/_default_template_yaml
+++ b/lmms_eval/tasks/covost2_zh_en/_default_template_yaml
@@ -1,0 +1,28 @@
+dataset_path: lmms-lab/covost2
+dataset_name: zh_en
+dataset_kwargs:
+  token: True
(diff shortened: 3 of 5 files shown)


lmms_eval/tasks/covost2_zh_en/covost2_zh_en_dev.yaml
```diff
@@ -0,0 +1,27 @@
...
+metric_list:
+  - metric: bleu
+    aggregation: !function utils.covost2_bleu
+    higher_is_better: false
+metadata:
+  - version: 0.0
+lmms_eval_specific_kwargs:
+  default:
+    pre_prompt: "Translate the speech into English and only output the translation: <|zh|>"
+    post_prompt: ""
```
kcz358: I think the `<|zh|>` symbol is a LID for the Qwen2-Audio. Possibly a too specific prompt for the default prompt. Consider change it and add as the Qwen2 Audio custom prompt.

lmms_eval/tasks/covost2_zh_en/covost2_zh_en_test.yaml
```diff
@@ -0,0 +1,27 @@
...
+metric_list:
+  - metric: bleu
+    aggregation: !function utils.covost2_bleu
+    higher_is_better: false
+metadata:
+  - version: 0.0
+lmms_eval_specific_kwargs:
+  default:
+    pre_prompt: "Translate the speech into English and only output the translation: <|zh|>"
+    post_prompt: ""
```
kcz358: Same question for this. And also, I think the two yaml files are very similar. Is it possible to combine them into a single `_default_yaml_template_zh_en` for convenience?

kcz358 (approved):
Hi, I think most of the parts are great, I think the only part is that whether we can use default template to reduce the replicate codes. Also, I think commit [58d8970](https://github.com/EvolvingLMMs-Lab/lmms-eval/pull/515/commits/58d8970e4ae0795e83db823222b29794389d611b) has been merged into main already. So would be good if you can drop that commit. I can help you with that if you don't know how to do it. Thank you!

pbcong:
Hi, thanks for the comment, i have fix the default prompt and added a default template. I also squashed all the commit to 1 commit. Can you kindly check and merge for me. Thank you!

kcz358 (approved):

PR #526, opened 2025-02-09
[Feat] add @maj and @pass to support sampling multiples times during evaluation

Luodian:
With this PR, we fixed two problems:

1. Display the group results for `mmlu_flan_n_shot_generative` by changing the `acc` to `exact_match` since `acc` will result the group display not matched in evaluation metrics.

image

2. Introduce the @maj and @pass mechanisms to enable majority voting and multi-pass evaluation. In this mode, the model generates multiple response samples for a given question. The final correctness is determined either through majority voting (@maj), where the most frequently occurring answer is selected as the final output, or through multi-pass evaluation (@pass), where the presence of at least one correct response across the generated samples is sufficient to deem the final output correct.

image

### DIFF
--- a/lmms_eval/api/metrics.py
+++ b/lmms_eval/api/metrics.py
@@ -1,5 +1,4 @@
 # the code is adapted from https://github.com/EleutherAI/lm-evaluation-harness
-import logging
 import math
 import random
 import re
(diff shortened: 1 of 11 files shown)


kcz358 (approved):
## Notice

Few shot context partially integrated

ghost:
Hi. I wonder how to use majority voting mode?

Luodian:
> Hi. I wonder how to use majority voting mode?

you can try to use "mmmu_val_pass64" like this task to enable majo vote.

PR #544, opened 2025-02-19
[Model] add vllm compatible models

Luodian:
Add vLLM model integration and update configurations
- Introduce VLLM model in the model registry.

vLLM offers exceptional speed in evaluation. Moving forward, we should prioritize using this approach.

image


lmms_eval/models/vllm.py
```diff
@@ -0,0 +1,179 @@
...
+NUM_SECONDS_TO_SLEEP = 5
+
+try:
+    import vllm
+except ImportError:
+    vllm = None
+
+
+@register_model("vllm")
+class VLLM(lmms):
```
kylesayrs, line 32: Since this processing is specific to VLMs, consider renaming to vllm_vlm

lmms_eval/models/vllm.py
```diff
@@ -0,0 +1,179 @@
...
+        if isinstance(image, str):
+            img = Image.open(image).convert("RGB")
+        else:
+            img = image.copy()
+
+        output_buffer = BytesIO()
+        img.save(output_buffer, format="PNG")
+        byte_data = output_buffer.getvalue()
+
+        base64_str = base64.b64encode(byte_data).decode("utf-8")
```
kylesayrs, line 91: Is base64 specific to qwen_vl? The vllm VLM interface accepts PIL images and then does model-specific processing

fzyzcjy:
+1 This looks great! 

I wonder whether this PR is ready for use for not? i.e. can I just use this branch and run tests

lmms_eval/models/vllm.py
Luodian: great, let me change it

lmms_eval/models/vllm.py
Luodian: I think this can remain as is `vllm` supports both vision/language models and vision/language tasks. For example, actually you can use `Qwen/Qwen2.5-0.5B-Instruct` to evaluate `mmlu_flan_n_shot_generative` within our framework.

So it's better supporting `vllm` through a single class.

Luodian:
> +1 This looks great and thanks for the support!
> 
> I wonder whether this PR is ready for use for not / when will this be ready for use? i.e. can I just use this branch and run it

think immediately

lmms_eval/models/vllm.py
Luodian: oh after checking the doc, I feel like we should keep the openai format messages so we dont need to apply chat template. 

Otherwise if we use this format, we need to apply model specific chat template.

```
outputs = llm.generate({
    "prompt": prompt,
    "multi_modal_data": {"image": image_embeds},
})
```

https://github.com/vllm-project/vllm/blob/0d243f2a54fbd1c56da8a571f0899c30b6aba5d9/docs/source/serving/multimodal_inputs.md

Luodian:
image

kcz358 (approved):

PR #558, opened 2025-02-26
Merging Ola

Devininthelab:

- [ ] A detailed description

```sh
pip install pre-commit
pre-commit install
pre-commit run --all-files
```

### DIFF
--- a/lmms_eval/models/ola.py
+++ b/lmms_eval/models/ola.py
@@ -264,7 +264,7 @@ class Ola(lmms):
 
         spare_frames = vr.get_batch(frame_idx).asnumpy()
         video = [PIL.Image.fromarray(frame) for frame in spare_frames]
-        return video
+        return video, frame_idx
 
     def tok_decode(self, tokens):
         return self.tokenizer.decode(tokens)
@@ -460,7 +460,7 @@ class Ola(lmms):
                         MODALITY = "VIDEO"
                     # Process audio of video
                     try: 
-                        video = self.load_video(visual, self.max_frames_num)
+                        video, frame_idx = self.load_video(visual, self.max_frames_num)
                     except Exception as e:
                         eval_logger.info(f"{e}")
                         eval_logger.info(f"Video {visuals} can not load, check the source")
--- a/lmms_eval/models/ola.py
+++ b/lmms_eval/models/ola.py
@@ -10,6 +10,7 @@ os.environ['MINRES'] = '0'
 os.environ['FORCE_NO_DOWNSAMPLE'] = '1'
 os.environ['LOAD_VISION_EARLY'] = '1'
 os.environ['PAD2STRIDE'] = '1'
+os.environ['USE_SPEECH'] = '1'
 import logging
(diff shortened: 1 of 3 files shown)


Devininthelab:
@kcz358  Hi Kaichen Zhang, please help review my code

lmms_eval/models/ola.py
```diff
@@ -0,0 +1,693 @@
...
+
+    The path for the project should be like this:
+        Folder/contains/lmms-eval/and/Ola  
+            ├── lmms-eval (current directory)  
+            │   ├── lmms_eval/  
+            │   ├── ...  
+            │   ├── large-v3.pt  
+            │   ├── BEATs_iter3_plus_AS2M_finetuned_on_AS2M_cpt2.pt  
+            ├── Ola  
+            │   ├── ...
```
kcz358, line 106: May I ask is this audio encoder path being hardcoded for Ola?

kcz358: I saw that in the Ola 7B, the beats encoder already in the safetensors. So can directly use load pretrained right?

kcz358 (commented):
Hi @Devininthelab , most of the code seems okay to me. Can you please fix the lint by running pre-commit. Then I will merge the code. Thanks!

lmms_eval/models/ola.py
Devininthelab: Hi I think for Ola in their repo, only the LLM are in safetensors. I checked through their [link](https://huggingface.co/THUdyh/Ola-7b/tree/main).

kcz358: I also checked their repo readme and see their instructions. But indeed in their safetensors, it contains speech encoder but not only the LLM. Does this mean that it needs be loaded twice?

image
image

Devininthelab:
> Hi @Devininthelab , most of the code seems okay to me. Can you please fix the lint by running pre-commit. Then I will merge the code. Thanks!

Hi I have fixed the Lint, please help me review again

YJCX330:
I have a question about the evaluation code for Ola. In lmms_eval/models/ola.py:  in lines 282-286, the video has already been sampled down to 64 frames, but in the loop at line 498, the video variable seems to be treated as the original video and is sampled again. After this loop ends, the video_processed variable does not contain 64 frames. Is this a bug?

Devininthelab:
@YJCX330 I will check again

Devininthelab:
@YJCX330 
Hi,

The logic of Ola running like this:

Inside the loop, when line [482](https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/main/lmms_eval/models/simple/ola.py#L482) is triggered, it then runs the function [load_video](https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/main/lmms_eval/models/simple/ola.py#L276-L286). It is from the function load video where the frames are sampled. 

For Ola, we input pure video_path. So that's why we have to sample frames like that to input to the model.

Hopes it answer your question.

PR #613, opened 2025-03-31
Add Omni Bench

ngquangtrung57:
Add Omni Bench with 3 settings:
- Original
- Replace audio with provided audio transcript ( for vlm)
- Replace image with provided image caption (for alm)
During evaluation, there are 12 samples that soundfile unable to load the audio.

I try to work around this but cannot fix. Hence I remove those samples and upload to lmms-lab/Omni_Bench_fix. The full set is at lmms-lab/Omni_Bench. The original has 1342 samples, the fixed one has 1330 samples.

### DIFF
--- a/lmms_eval/tasks/omni_bench/utils.py
+++ b/lmms_eval/tasks/omni_bench/utils.py
@@ -1,8 +1,10 @@
-import random
-import numpy as np
-from collections import defaultdict
 import logging
+import random
+from collections import defaultdict
+
+import numpy as np
 from loguru import logger as eval_logger
+
 
 def omni_bench_original_doc_to_visual(doc):
     return [doc["audio"], doc["image"]]
@@ -15,13 +17,9 @@ def omni_bench_original_doc_to_text(doc, lmms_eval_specific_kwargs):
     options = doc["options"]
     letters = ["A", "B", "C", "D"]
     enumerated_opts = [f"{letters[i]}. {opt}" for i, opt in enumerate(options)]
-    prompt_text = (
-        f"{pre_prompt}"
-        f"Question:\n{question}\n\n"
-        f"Options:\n" + "\n".join(enumerated_opts) +
-        f"\n\n{post_prompt}"
-    )
+    prompt_text = f"{pre_prompt}" f"Question:\n{question}\n\n" f"Options:\n" + "\n".join(enumerated_opts) + f"\n\n{post_prompt}"
     return prompt_text
+
 
 def omni_bench_image_caption_doc_to_visual(doc):
     return [doc["audio"]]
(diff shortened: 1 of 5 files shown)


lmms_eval/tasks/omni_bench/utils.py
```diff
@@ -0,0 +1,138 @@
...
+    question = doc["question"]
+    options = doc["options"]
+    letters = ["A", "B", "C", "D"]
+    enumerated_opts = [f"{letters[i]}. {opt}" for i, opt in enumerate(options)]
+    prompt_text = f"{pre_prompt}" f"Question:\n{question}\n\n" f"Options:\n" + "\n".join(enumerated_opts) + f"\n\n{post_prompt}"
+    return prompt_text
+
+
+def omni_bench_image_caption_doc_to_visual(doc):
+    return [doc["audio"]]
```
kcz358, line 25: Is this audio or image?

lmms_eval/tasks/omni_bench/utils.py
```diff
@@ -0,0 +1,138 @@
...
+    question = doc["question"]
+    options = doc["options"]
+    letters = ["A", "B", "C", "D"]
+    enumerated_opts = [f"{letters[i]}. {opt}" for i, opt in enumerate(options)]
+    prompt_text = f"{pre_prompt}" f"Visual Context for the audio:\n{image_caption}\n\n" f"Question:\n{question}\n\n" f"Options:\n" + "\n".join(enumerated_opts) + "\n\n" f"{post_prompt}"
+    return prompt_text
+
+
+def omni_bench_audio_transcript_doc_to_visual(doc):
+    return [doc["image"]]
```
kcz358, line 41: Is this audio or image?

kcz358 (commented):
I am not sure whether my understanding is correct. So is it like omnibench provide image caption for audio testing and audio transcript for image testing?

lmms_eval/tasks/omni_bench/utils.py
ngquangtrung57: this is audio. image caption means the image replace with image caption, so doc_to_visual includes only "audio"

lmms_eval/tasks/omni_bench/utils.py
ngquangtrung57: this is image. audio transcript means audio replace with audio transcript, so doc_to_visual includes only "image"

ngquangtrung57:
yes, for Audio Language Model, it provides audio and image caption as context (because ALM cannot understand image). And for VLM, it provides image and audio transcript as context (because VLM cannot understand audio). So this benchmark can work for not only Omni models but also VLMs and ALMs.

kcz358 (approved):

PR #584, opened 2025-03-17
Add support for specifying local dataset

harrytea:
### Summary

This PR improves the flexibility of specifying local datasets, addressing issues where it was inconvenient to do so.

### Changes Made

1.	Modified api.task.py
	 •	Added local_dataset_path to support loading datasets from a local directory.
2.	Updated YAML Files in tasks/
	•	Added local_dataset_path under dataset_path.
	•	If specified, data is loaded from os.path.join(local_dataset_path, dataset_path); otherwise, it is downloaded from Hugging Face.
3.	Added local_infer_dataset_path
	•	Applied the same logic as local_dataset_path for inference datasets.
	•	Updated relevant files (e.g., tasks.gqa.utils.py) to support this option.

PR #629, opened 2025-04-12
Fix loading datasets from disk

CLARKBENHAM:
If you want to load a dataset from a local location there were 2 bugs: 
- the kwargs you needed for a later call would get popped since a copy isn't made of the kwargs
-  Params for datasets.load_dataset were used instead of datasets.load_from_disk
Related to https://github.com/EvolvingLMMs-Lab/lmms-eval/pull/584


Luodian:
Thanks for supporting LMMs-Eval!

PR #641, opened 2025-04-18
[Fix] Aria and LLama Vision and OpenAI compatible models

Luodian:
- Replaced AutoModelForCausalLM and AutoProcessor with AriaForConditionalGeneration and AriaProcessor in the Aria model.
- Updated the pretrained model string in LlamaVision to "meta-llama/Llama-3.2-11B-Vision-Instruct".
- Enhanced OpenAICompatible to support AzureOpenAI and modified API key handling for better flexibility.
- Adjusted timeout parameter and refined token handling in OpenAICompatible for improved functionality.

- [ ] A detailed description

```sh
pip install pre-commit
pre-commit install
pre-commit run --all-files
```

### DIFF
--- a/lmms_eval/models/aria.py
+++ b/lmms_eval/models/aria.py
@@ -10,7 +10,7 @@ from accelerate.state import AcceleratorState
 from decord import VideoReader, cpu
 from PIL import Image
 from tqdm import tqdm
-from transformers import AutoModelForCausalLM, AutoProcessor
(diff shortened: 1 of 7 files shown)


kcz358 (approved):
Most of the codes look fine to me as this is the only way to hardcode under current logic. I will merge this PR.

A small notice for others:

We plan to rfc the doc to text these things in the future with messages format so these hardcoded models will gradually being deprecated. Please stay tuned for our update

Issue #648, opened 2025-04-22
Bug: The images blending within the same batch leads to the performance degradation of Qwen2.5_VL when the batch size increases


ashun989:
# Environment
```
lmms-eval=0.3.3
transformers=4.51.3
```

# Context

I used the following commands to test the performance of the Qwen2.5VL-3B model on the `vqav2_val_lite` task, with batch sizes of 1, 2, and 4 respectively.

```shell
task="vqav2_val_lite"

CUDA_VISIBLE_DEVICES=0 \
accelerate launch --num_processes=1 --main_process_port=12345 -m lmms_eval \
    --model qwen2_5_vl \
    --model_args=pretrained=./Qwen/Qwen2.5-VL-3B-Instruct,use_flash_attention_2=True \
    --tasks $task \
    --batch_size 1 \
    --output_path ./result/Qwen2.5-VL-3B-Instruct/lmms_eval_$task \
    --log_samples

CUDA_VISIBLE_DEVICES=0 \
accelerate launch --num_processes=1 --main_process_port=12345 -m lmms_eval \
    --model qwen2_5_vl \
    --model_args=pretrained=./Qwen/Qwen2.5-VL-3B-Instruct,use_flash_attention_2=True \
    --tasks $task \
    --batch_size 2 \
    --output_path ./result/Qwen2.5-VL-3B-Instruct/lmms_eval_${task}_bs2 \
    --log_samples


CUDA_VISIBLE_DEVICES=0 \
accelerate launch --num_processes=1 --main_process_port=12345 -m lmms_eval \
    --model qwen2_5_vl \
    --model_args=pretrained=./Qwen/Qwen2.5-VL-3B-Instruct,use_flash_attention_2=True \
    --tasks $task \
    --batch_size 4 \
    --output_path ./result/Qwen2.5-VL-3B-Instruct/lmms_eval_${task}_bs4 \
    --log_samples
```

The results are as follows:

|Batch size| Performance |
|:---:|:---:|
| 1 | 0.7478 |
| 2 | 0.7234 |
| 4 | 0.6810 |

# Observation

Based on my current observations, the issue might be located [here](https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/514082ea326d903f7dfed9ec04bdbc70b7018015/lmms_eval/models/qwen2_5_vl.py#L192-L278)

Simply put, the situation in the code seems to be like this:

```python
# `contexts` is a batch of context messages.
[...]

PR #649, opened 2025-04-22
Fix the bug in issue #648

ashun989:
Fix the bug: The images blending within the same batch leads to the performance degradation of Qwen2.5_VL when the batch size increases

After fixing:

- model: Qwen/Qwen2.5-VL-3B-Instruct
- task: vqav2_val_lite
- results: 

| Batch size | Performance | Total Time(s) |
|:---:|:---:|:---:|
| 1 | 0.7478 | 120.2 |
| 2 | 0.7490 | 109.9 |
| 4 | 0.7466 | 98.9 |


Luodian (approved):
quite nice, thank you!

PR #656, opened 2025-04-27
[Task] Add new benchmark: CAPability

lntzm:
This pull request adds a new visual caption benchmark: CAPability.

[CAPability](https://github.com/ali-vilab/CAPability) evaluates the LMMs by around 11K human-annotated image/video-annotation pairs. Different from directly annotating ground-truth captions, we design 13 dimensions and annotate information for each dimension, and evaluate from these 13 dimensions, calculating precision, recall, and f1-score as metrics. See this [README](https://github.com/ali-vilab/CAPability/blob/main/README.md) for detailed introduction. Therefore, we create 1 default template yaml, 13 config yamls for each dimension, and 1 final grouped config yaml. The `utils.py` and `prompt.py` support the inference and evaluation process.

Check these links if you are interested in CAPability: [ArXiv](https://arxiv.org/abs/2502.14914), [Project Page](https://capability-bench.github.io/), [Huggingface Dataset](https://huggingface.co/datasets/lntzm/CAPability), and [Github Repo](https://github.com/ali-vilab/CAPability).

This is an example of an evaluation pipeline:
example

This PR does not break any existing functionality in lmms-eval.

### DIFF
--- a/lmms_eval/tasks/capability/utils.py
+++ b/lmms_eval/tasks/capability/utils.py
@@ -114,6 +114,7 @@ def capability_aggregate_inference_result(results, args):
     # to ensure re-run evaluation if re-run inference
     eval_save_path = os.path.join(os.path.dirname(save_path), f"../evaluation/{task}.jsonl")
     if os.path.exists(eval_save_path):
+        eval_logger.warning(f"Found EXISTING evaluation records: {eval_save_path}, REMOVING it!")
         os.remove(eval_save_path)
     
     with open(save_path, 'w') as f:
--- a/lmms_eval/tasks/capability/utils.py
+++ b/lmms_eval/tasks/capability/utils.py
@@ -1,0 +1,622 @@
+import os
+import ast
+import yaml
+import json
+import numpy as np
+from PIL import Image
+from tqdm import tqdm
+from pathlib import Path
+from openai import OpenAI
+from loguru import logger as eval_logger
+from lmms_eval.tasks.capability.prompt import Prompts
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from lmms_eval.tasks._task_utils.file_utils import generate_submission_file
+
+
+with open(Path(__file__).parent / "_default_template_yaml", "r") as f:
+    raw_data = f.readlines()
+    safe_data = []
+    for i, line in enumerate(raw_data):
+        # remove function definition since yaml load cannot handle it
+        if "!function" not in line:
+            safe_data.append(line)
+config = yaml.safe_load("".join(safe_data))
+
+OPENAI_API_URL = os.getenv("OPENAI_API_URL", "https://api.openai.com/v1/chat/completions")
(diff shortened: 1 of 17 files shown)


lmms_eval/tasks/capability/utils.py
```diff
@@ -0,0 +1,622 @@
...
+    raw_data = f.readlines()
+    safe_data = []
+    for i, line in enumerate(raw_data):
+        # remove function definition since yaml load cannot handle it
+        if "!function" not in line:
+            safe_data.append(line)
+config = yaml.safe_load("".join(safe_data))
+
+OPENAI_API_URL = os.getenv("OPENAI_API_URL", "https://api.openai.com/v1/chat/completions")
+OPENAI_API_KEY = os.getenv("OPENAI_API_KEY", "YOUR_API_KEY")
```
kcz358: For consistency, it is better to also support loading from Azure endpoint also. You can check other tasks such as llava in the wild for reference

lmms_eval/tasks/capability/utils.py
```diff
@@ -0,0 +1,622 @@
...
+            file_name=f"{task}.jsonl",
+            args=args,
+            subpath=f"capability_results/{suffix}/inference"
+        )
+
+    # delete the invalid evaluation results as lmms-eval do not support auto-resume inference
+    # to ensure re-run evaluation if re-run inference
+    eval_save_path = os.path.join(os.path.dirname(save_path), f"../evaluation/{task}.jsonl")
+    if os.path.exists(eval_save_path):
+        os.remove(eval_save_path)
```
kcz358: Should user also pay attention to this hardcoded eval save path?

kcz358 (approved):
Hi, thank you for your contribution. Most of the codes LGTM, you may check some of the comments to see whether it should be addressed. Otherwise, I will merge this PR. Thanks!

lmms_eval/tasks/capability/utils.py
lntzm: This hardcoded save path is used for saving the evaluation result when running the first metric (precision), and then directly reading the record when running the second and third metrics (recall and f1-score) rather than re-running GPT evaluation. I am not sure whether there is a better way to achieve this. 

The removal of this path is in the case of unexpected interruption. As lmms-eval does not support resuming eval, user has to re-run the inference and evaluation when meeting interruption. Then the previously saved evaluation records should be removed to avoid loading it. What is expected to do to let users pay attention to this hardcoded eval save path? Maybe add a config to control?

kcz358: No problem about hardcoding this right now. I think currently you can add a warning log here for to notify the user about this. I may add the resume evaluation into the next upgrade feature if possible

lntzm: get it~

lntzm:
Hi~ I have updated 2 more commits to fit the suggestion. Thank you for checking the PR, and please feel free to let me know if there are any other suggestions for improving the code.

kcz358 (approved):
Hi, LGTM. Thank you for your contributions!

PR #690, opened 2025-05-27
Pass in the 'cache_dir' to use local cache

JulyanZhu:
Hi, thank you for your awesome project.
I recently wanted to modify the yaml file to use my local dataset since I already have them on my machine.
So I set the 'cache_dir' in the 'dataset_kwargs' and noticed that the lmms-eval is still trying to fetch dataset from huggingface instead of using the local one.
After throughly check, I noticed we can pass in the 'cache_dir' to the huggingface snapshot_download function in the tasks.py if we have 'videos' in dataset_kwargs to avoid force-redownloading.


kcz358:
I'm gonna revert this PR for now. From testing, this will cause people that want to extract and download videos from the hub to be unable to do so. Because after snapshot download, the existing cache dir will cause unzipping logic to fail. Later we may give an option to skip snapshot download and load from cache directly, but I will revert this PR for now.

PR #727, opened 2025-06-23
add new tasks MMVU and Visual Web Bench

pbcong:
add new tasks:
- MMVU
- Visual Web Bench


Luodian:
Hi thanks @pbcong can you address the conflicts and then we can merge to `dev/0d4`

Issue #712, opened 2025-06-11
[Bug?] Doc fields with "image" not saved to jsonl file when using log_samples=True


YongchengYAO:
When setting `log_samples=True`, the doc, target, model output, and other info are saved to a jsonl file.

With this line:

https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/d4383329aeaa6ffbcde94a9b31ca0eff7fee557c/lmms_eval/evaluator.py#L525

in the `evaluator.py`, any fields with "image" in the doc key are filtered out. Any reason for that?

Code block:
```python
if log_samples:
      target = task.doc_to_target(doc)
      saved_doc = {}
      for key, value in doc.items():
          # If image is not in key
          if "image" not in key:
              # If audio is also not the value
              if isinstance(value, dict) and "array" in value:
                  continue
              else:
                  saved_doc[key] = value
      filtered_arguments = []
      for req in requests:
          # check if req.args is a list of tuples, and each item in the list is a serializable object
          for value in req.args:
              if isinstance(value, (str, int, float, bool, list, dict, type(None))):
                  filtered_arguments.append(value)
              # else:
              #     filtered_arguments.append(_handle_non_serializable(value))

      example = {
          "doc_id": doc_id,
          "doc": saved_doc,
          "target": target,
          "arguments": filtered_arguments,
          "resps": [req.resps for req in requests],
          "filtered_resps": [req.filtered_resps[filter_key] for req in requests],
          "doc_hash": hash_string(
              json.dumps(
                  requests[0].doc,
                  indent=2,
                  default=handle_non_serializable,
                  ensure_ascii=False,
              )
          ),
          "prompt_hash": hash_string(requests[0].arguments[0]),
          "target_hash": hash_string(str(target)),
      }
      example.update(metrics)
      task_output.logged_samples.append(example)
```

Luodian:
Yes actually in our original design, the image is not supposed to be saved since it can be viewed in the huggingface. Do you think it should be saved locally? Wish to see your thoughts and design on this feature!

YongchengYAO:
I think it is valuable to keep the record that links images to questions. The most useful usecase of the log files is to parse the model responses. In some cases, users may want to check the input image for debugging or visualization. In my case, my HF dataset would download the image files and meta data containing the relative path of images. The dataset also returns meta info such as image size and other labels that may be used in model responses parsing.

[...]

PR #731, opened 2025-06-27
bugfix: missing fields in doc when using --log_samples

VincentYCYao:
Issue: https://github.com/EvolvingLMMs-Lab/lmms-eval/issues/712#issue-3137285923

Bug ❓: Fields with "image" in keys or of type dict are not saved in the sample log file.

Fix: save dicts and all fields to the  JSONL file

Testing: The update code has been tested with image-text-to-text inference. All fields including dicts in doc are saved in the log file.

## Summary by CodeRabbit

* **Bug Fixes**
  * All key-value pairs in evaluation samples are now saved without exclusion, ensuring that previously filtered data (such as images and audio arrays) are retained in the results.

### DIFF
--- a/lmms_eval/tasks/camerabench_vqa/utils.py
+++ b/lmms_eval/tasks/camerabench_vqa/utils.py
@@ -6,7 +6,6 @@ dir_name = os.path.dirname(os.path.abspath(__file__))
 SUFFIX_FOR_VQA = {"yes_no": "Please answer Yes or No.", "multiple_choice": "Please output the letter corresponding to the correct option."}
 
 
-
 def get_scores(scores):
     """
     Calculate various scores based on the given results.
@@ -148,11 +147,8 @@ def extract_answer(output_string, task_type="yes_no"):
 
 def cambench_doc_to_visual(doc):
     try:
-        default_path = os.path.join(os.getenv('HOME'), '.cache/huggingface')
-        load_path = os.path.expanduser(os.path.join(
-            os.getenv("HF_HOME", default_path),
-            'camerabench_vqa/datasets--chancharikm--camerabench_vqa_lmms_eval/snapshots'
-        ))
+        default_path = os.path.join(os.getenv("HOME"), ".cache/huggingface")
+        load_path = os.path.expanduser(os.path.join(os.getenv("HF_HOME", default_path), "camerabench_vqa/datasets--chancharikm--camerabench_vqa_lmms_eval/snapshots"))
 
         if not os.path.exists(load_path):
             raise FileNotFoundError(f"Dataset path not found: {load_path}")


VincentYCYao:
The reason to keep these fields with "image" is that people may have "image_size", "image_description", or "*image*":[dict] in the doc that will be used for metric calculation.

Note that dictionary is not saved in the current latest version.

kcz358:
I think in the later code

https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/e24a7d8ecd2d1c29ded2ac4b85167bc94e6d52ff/lmms_eval/evaluator.py#L540-L556

the saved doc will be added to the examples and will be added to the save doc when log samples is being used?

May I ask where will this being used for metric calculation? I think the save doc is after the process result step.

YongchengYAO:
Yes, the structure of the log file is like this:
```python
example = {
                        "doc_id": doc_id,
                        "doc": saved_doc,
                        "target": target,
                        "arguments": filtered_arguments,
                        "resps": [req.resps for req in requests],
                        "filtered_resps": [req.filtered_resps[filter_key] for req in requests],
                        "doc_hash": hash_string(
                            json.dumps(
                                requests[0].doc,
                                indent=2,
                                default=handle_non_serializable,
                                ensure_ascii=False,
                            )
                        ),
                        "prompt_hash": hash_string(requests[0].arguments[0]),
                        "target_hash": hash_string(str(target)),
                    }
```
where `saved_doc` contains the original info returned from a HF dataset. For instance, there would be {"doc": [RAW-doc-dict]} in the log file, where <RAW-doc-dict> could be like {"image_path": [some path], "image_metainfo": [some info], "key w/o the string image": [sth]}. 

The problem is that we are filtering the [RAW-doc-dict>] such that fields like "image_path" and "image_metainfo" are not saved in the log file. For example, if I have a custom HF dataset with complicated data structure and it comes with a data loading script (in HF dataset repo) that downloads the image data and returns the "image_path" when loaded with `load_dataset()`, the "image_path" is used to locate the image files.

Why it is desirable to keep the integrity of the raw doc info?
- People may want to find (or plot) the corresponding image for debugging purposes or to investigate model failure mode.
- Runing a model inference is expensive especially for a large dataset. And the model evaluation metrics are not always perfect when running the evaluation script. It is very likely that users may want to parse the model prediction after running the evaluation using additional post-processing script. In that case, some values in [RAW-doc-dict] could be useful for metric calculation.

kcz358:
The concern here is that in many many datasets currently integrated by lmms-eval, even the text part of the previous saved doc contains lots of content. When we performing evaluation, the saved doc would causing the saving results to be extreme slow and requires 200+ Mb to store the file. I would suggest this to be an optional choice to save full doc instead of hardcode it here

YongchengYAO:
I agree. It would be nice to have an argument controling this behavior and increase the transparency of this process. In my case, I discovered this issue after a few runs.

Thank you very much!

PR #747, opened 2025-07-09
Feature/inference throughput logging

Luodian:

- [ ] A detailed description

```sh
pip install pre-commit
pre-commit install
pre-commit run --all-files
```

### DIFF
--- a/docs/throughput_metrics.md
+++ b/docs/throughput_metrics.md
@@ -1,0 +1,123 @@
+# Throughput Metrics Documentation
+
+This document describes the inference throughput metrics automatically logged by LMMS-Eval chat models during evaluation.
+
+## Overview
+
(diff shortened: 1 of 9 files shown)


kcz358 (commented):
I think the idea and most of the implementation looks okay. But the format seems to be inconsistent for each model. I would see if I can create a context manager for this or fix the algorithm format. Also, are `TTFT` and `TPOT` necessary in our current settings? I think currently are all estimation so I feel this is a bit inaccurate and maybe tokens per second would be enough to log everything

Luodian:
> I think the idea and most of the implementation looks okay. But the format seems to be inconsistent for each model. I would see if I can create a context manager for this or fix the algorithm format. Also, are `TTFT` and `TPOT` necessary in our current settings? I think currently are all estimation so I feel this is a bit inaccurate and maybe tokens per second would be enough to log everything

Let me check, I think we should only implement this in chat models as standard guidance to community developers. So dont necessarily introduce more abstraction. I think TTFT/TPOT is from vllm/sglang's metrics so I keep it, let me dig deeper or you could start commit to revise it.

PR #813, opened 2025-09-03
Add lemonade benchmark to the evaluation

andybonnetto:
## Description 
We added the evaluation of Lemonade [(on HuggingFace)](https://huggingface.co/datasets/amathislab/LEMONADE). A motion centered MCQ dataset extracted from the EPFL-Smart-Kitchen-30 dataset ([ArXiv](https://arxiv.org/abs/2506.01608) and [Zenodo](https://zenodo.org/records/15535461)). The benchmark was evaluated using Lmms-eval in the publication.

## Modifications
* Added `lemonade` folder in tasks containing `lemonade.yaml` and `utils.py`
* `utils.py` contains data loading from [hugging face ](https://huggingface.co/datasets/amathislab/LEMONADE), video loading, answer parsing, metric evaluation.
* `lemonade.yaml` contains the default prompts

## Summary by CodeRabbit

* **New Features**
  * Added support for the LEMONADE multimodal evaluation task with a ready-to-use evaluation configuration.
  * Automatic download/extraction and caching of required video assets on first run.
  * Frame sampling from videos for consistent visual inputs.
  * Prompt templates and generation settings for deterministic multiple-choice outputs.
  * Robust answer parsing, structured result processing, and aggregated accuracy reporting.


lmms_eval/tasks/lemonade/utils.py
```diff
@@ -0,0 +1,271 @@
...
+
+LEMONADE_ZIP_NAMES = [
+    "videos_batch_0.zip",
+    "videos_batch_1.zip",
+    "videos_batch_2.zip",
+    "videos_batch_3.zip",
+    "videos_batch_4.zip",
+]
+
+data_dir = "./data/lemonade"
```
kcz358: The data path for video evaluation should not be hardcoded

lmms_eval/tasks/lemonade/utils.py
```diff
@@ -0,0 +1,271 @@
...
+    video_path = os.path.join(videos_dir, video_filename)
+
+    if os.path.exists(video_path):
+        start = int(doc["Start"])
+        end = int(doc["End"])
+        frames = load_video(video_path, start, end, max_num_frames=MAX_NUM_FRAMES)
+    else:
+        raise FileNotFoundError(f"Video file not found: {video_path}")
+
+    return frames
```
kcz358: The doc to visual of video should not be frames. Instead, it should return the list of video path

andybonnetto: Due to the long duration of the videos and the repetitiveness/overlap of the video segments, it is wiser to avoid chunking the video prior to access to avoid memory redundancy. This framework (outputting frames) was inspired by another task pushed in your repository : https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/main/lmms_eval/tasks/plm_videobench/rcap/rcap_utils.py.

kcz358: It would be okay if you always want to hack your video frames for every model to test and we would accept merge like that. But we would suggest you to return a video path instead of video frames as multiple images so that the video chunking can be done in the model generation itself (for example using the general api we provided) so that fps and max frames can be set by the user.

kcz358:
Hi, thank you for the PR. I think for the major issue is that we hope that your download data can be placed in the snapshot cache and the cache folder can be always settable using HF_HOME instead of directly setting `./data` with a relative path. Can you please check other video tasks to see how to achieve that? Thanks!

TashkovskaMatea:
Hi, thank you for the feedback. We've updated the code so that the download data is now placed in the snapshot cache, following the same approach used in other video tasks. Looking forward to your feedback on this update.

lmms_eval/tasks/lemonade/utils.py
```diff
@@ -0,0 +1,356 @@
...
+            filename=zip_name,
+            repo_type="dataset",
+            local_dir=os.path.join(base_cache_dir, "lemonade_zips"),
+            local_dir_use_symlinks=False,
+            resume_download=True,
+        )
+        with zipfile.ZipFile(zip_path, "r") as zf:
+            zf.extractall(videos_dir)
+
+    print("All videos downloaded and extracted successfully.\n")
```
kcz358: I would suggest this should not be included. As the download part is being done in the `api/task.py`. We handle the snapshot download from the hub.

lmms_eval/tasks/lemonade/utils.py
```diff
@@ -0,0 +1,356 @@
...
+import cv2
+import numpy as np
+import zipfile
+from collections import defaultdict
+from PIL import Image
+from typing import Any, Optional
+from huggingface_hub import hf_hub_download
+
+MAX_NUM_FRAMES = 8
+LEMONADE_ZIP_NAMES = [f"videos_batch_{i}.zip" for i in range(5)]
```
kcz358: Suggest not hardcoding this. Otherwise would be really strange for a video task to use only 8 frames or is this intended?

kcz358 (commented):
Hi, I would say it would be fine if you want to hardcode the download logic and load video logic for your benchmarks since it would not disturb the overall pipeline of the framework. But I would again suggest to leave these functionality to be handle on other components of the PR.

Feel free to add a README in your dataset folder on how to use or any comments on the implementation. If you feel good for all of your changes, I will merge it.

lmms_eval/tasks/lemonade/utils.py
TashkovskaMatea: Thank you for the suggestion, we removed the custom download function. The dataset now relies on the existing download/unzip logic in api/task.py.

lmms_eval/tasks/lemonade/utils.py
TashkovskaMatea: We removed the hardcoded value and the max_num_frames is now configurable through the task YAML. We can’t set this at model generation time since frame extraction happens during the video loading (due to the long duration of the videos and the overlap/repetitiveness of segments).

TashkovskaMatea:
Thanks again for the helpful suggestions - we’ve made the changes and added a README. All set for merge on our side!

kcz358 (approved):

PR #856, opened 2025-10-12
fix(qwen_vl_api): Fix temporary files being deleted prematurely for multi-image inputs

LeeeeTX:
### 📝 Description

This PR fixes a bug in the `qwen_vl_api` model wrapper where requests with multiple images would fail because temporary image files were being deleted prematurely.

### 🐛 The Bug

The previous implementation used `tempfile.NamedTemporaryFile(delete=True)` to create a temporary file for each image within a loop. The variable holding the file object (`temp_file`) was reassigned in each iteration.

This caused the file objects from previous iterations to lose all references, making them eligible for Python's garbage collection. When the garbage collector reclaimed an object, its file handle was closed, and because `delete=True` was set, the underlying file was immediately deleted from the disk.

As a result, only the temporary file for the very last image in the loop would persist, leading to "file not found" errors when the API call was made.

### ✅ The Fix

This fix addresses the issue by implementing the following changes:

1. **Preserving Object Lifecycles**: A list (`temp_files`) is introduced to hold references to all `NamedTemporaryFile`objects created during the loop. This prevents them from being garbage collected until they are no longer needed.
2. **Ensuring Resource Cleanup**: The logic for image processing and the API call is now wrapped in a `try...finally`block. The `finally` block ensures that `.close()` is called on every file object in the `temp_files` list, regardless of whether the API call succeeded or failed. This guarantees that all temporary files are properly cleaned up, preventing file leaks.

This fix ensures the stability and reliability of the model when handling complex, multi-image inputs.

### DIFF
--- a/lmms_eval/models/simple/qwen_vl_api.py
+++ b/lmms_eval/models/simple/qwen_vl_api.py
@@ -91,60 +91,68 @@ class Qwen_VL_API(lmms):
             # encode, pad, and truncate contexts for this batch
             visuals = [doc_to_visual(self.task_dict[task][split][doc_id])]
             visuals = self.flatten(visuals)
-            imgs = []
 
-            for idx, visual in enumerate(visuals):
-                temp_file = self.save_image_to_temp_file(visual)
-                imgs.append(temp_file.name)
-
-            messages = [{"role": "user", "content": []}]
-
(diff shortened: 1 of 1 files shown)


Luodian:
@claude

kcz358 (approved):

PR #861, opened 2025-10-17
feat: Better video tokens and prompt alignment with sglang and add vllm generate

kcz358:

- [ ] A detailed description

```sh
pip install pre-commit
pre-commit install
pre-commit run --all-files
```


kcz358:
This PR is used to add an option to first prepare prompts then add the mm data so that for models such as Qwen3-VL can be prepared with correct video timestamps

PR #876, opened 2025-10-26
[TASK] Added MindCube Task

oscarqjh:
added MindCube task from https://github.com/mll-lab-nu/MindCube


oscarqjh:
Hi, I have added a default yaml as suggested. Please take a look. Thank you!

PR #883, opened 2025-11-04
Add Qwen3-VL models (w/o vllm or sglang)

ArdalanM:
Inference Qwen3-VL models only using huggingface backend (no vllm/sglang).
Useful to better debug benchmark regressions

requires: transformers==4.57.1

Example results using Qwen3-VL-4B-Instruct
```
batch_size=1
qwen3_vl (pretrained=Qwen/Qwen3-VL-4B-Instruct,max_pixels=12845056,attn_implementation=flash_attention_2,interleave_visuals=False), gen_kwargs: (), limit: None, num_fewshot: None, batch_size: 1
|    Tasks     |Version|     Filter     |n-shot|        Metric         |   | Value |   |Stderr|
|--------------|-------|----------------|-----:|-----------------------|---|-------|---|------|
|mmbench_en_dev|Yaml   |none            |     0|gpt_eval_score         |↑  |84.1924|±  |   N/A|
|mmmu_val      |      0|none            |     0|mmmu_acc               |↑  | 0.5022|±  |   N/A|
|mmstar        |Yaml   |none            |     0|average                |↑  | 0.5832|±  |   N/A|
|ocrbench      |Yaml   |none            |     0|ocrbench_accuracy      |↑  | 0.8570|±  |   N/A|
|realworldqa   |Yaml   |flexible-extract|     0|exact_match            |↑  | 0.7098|±  |0.0164|

batch_size=8
qwen3_vl (pretrained=Qwen/Qwen3-VL-4B-Instruct,max_pixels=12845056,attn_implementation=flash_attention_2,interleave_visuals=False), gen_kwargs: (), limit: None, num_fewshot: None, batch_size: 8
|    Tasks     |Version|     Filter     |n-shot|        Metric         |   | Value |   |Stderr|
|--------------|-------|----------------|-----:|-----------------------|---|-------|---|------|
|mmbench_en_dev|Yaml   |none            |     0|gpt_eval_score         |↑  |84.1924|±  |   N/A|
|mmmu_val      |      0|none            |     0|mmmu_acc               |↑  | 0.4978|±  |   N/A|
|mmstar        |Yaml   |none            |     0|average                |↑  | 0.5832|±  |   N/A|
|ocrbench      |Yaml   |none            |     0|ocrbench_accuracy      |↑  | 0.8620|±  |   N/A|
|realworldqa   |Yaml   |flexible-extract|     0|exact_match            |↑  | 0.7098|±  |0.0164|
```

### DIFF
--- a/lmms_eval/models/chat/qwen3_vl.py
+++ b/lmms_eval/models/chat/qwen3_vl.py
@@ -66,7 +66,7 @@ class Qwen3_VL(Qwen3_VLSimple):
             else:
                 video_kwargs["nframes"] = self.max_num_frames
             batched_messages = [chat_message.to_hf_messages(video_kwargs=video_kwargs) for chat_message in chat_messages]
-            texts = [self.processor.apply_chat_template(msg, tokenize=False, add_generation_prompt=True) for msg in batched_messages]
+            texts = self.processor.apply_chat_template(batched_messages, tokenize=False, add_generation_prompt=True)
             image_inputs, video_inputs = process_vision_info(batched_messages)
             if video_inputs is not None:
                 total_frames = video_inputs[0].shape[0]
@@ -75,7 +75,8 @@ class Qwen3_VL(Qwen3_VLSimple):
                 if total_frames - 1 not in indices:
                     indices = np.append(indices, total_frames - 1)
                 video_inputs[0] = video_inputs[0][indices]
-            inputs = self.processor(text=texts, images=image_inputs, videos=video_inputs, padding=True, return_tensors="pt")
+            padding_side = "left" if self.batch_size > 1 else "right"
+            inputs = self.processor(text=texts, images=image_inputs, videos=video_inputs, padding=True, padding_side=padding_side, return_tensors="pt")
 
             if self.device_map == "auto":
                 inputs = inputs.to("cuda")
--- a/lmms_eval/models/simple/qwen3_vl.py
+++ b/lmms_eval/models/simple/qwen3_vl.py
@@ -14,7 +14,7 @@ from transformers import (
     AutoProcessor,
     AutoTokenizer,
     Qwen3VLForConditionalGeneration,
-    Qwen3VLMoeForConditionalGeneration
+    Qwen3VLMoeForConditionalGeneration,
 )
 
 from lmms_eval import utils


lmms_eval/models/chat/qwen3_vl.py
```diff
@@ -0,0 +1,152 @@
...
+                "max_pixels": self.max_pixels,
+                "min_pixels": self.min_pixels,
+            }
+            if self.fps is not None:
+                video_kwargs["fps"] = self.fps
+            else:
+                video_kwargs["nframes"] = self.max_num_frames
+            batched_messages = [chat_message.to_hf_messages(video_kwargs=video_kwargs) for chat_message in chat_messages]
+            texts = [self.processor.apply_chat_template(msg, tokenize=False, add_generation_prompt=True) for msg in batched_messages]
+            image_inputs, video_inputs = process_vision_info(batched_messages)
```
kcz358, line 70: When process batch messages, I think a direct apply chat template to the batched_messages could work?

lmms_eval/models/chat/qwen3_vl.py
```diff
@@ -0,0 +1,152 @@
...
+            if video_inputs is not None:
+                total_frames = video_inputs[0].shape[0]
+                indices = np.linspace(0, total_frames - 1, self.max_num_frames, dtype=int)
+                # Append the last frame index if not already included
+                if total_frames - 1 not in indices:
+                    indices = np.append(indices, total_frames - 1)
+                video_inputs[0] = video_inputs[0][indices]
+            inputs = self.processor(text=texts, images=image_inputs, videos=video_inputs, padding=True, return_tensors="pt")
```
kcz358: Might need to set padding side to left when process batched dataset to tokenizer.

kcz358:
Hi thank you for the contribution, I think most of the code LGTM. Just wondering if there are some small detail that might cause problem when bs > 1. Thanks!

lmms_eval/models/chat/qwen3_vl.py
ArdalanM: Yes it works, thanks !

lmms_eval/models/chat/qwen3_vl.py
ArdalanM: Yes, I updated the code, thanks !

lmms_eval/models/chat/qwen3_vl.py
```diff
@@ -75,7 +75,8 @@ def _collate(x):
                 if total_frames - 1 not in indices:
                     indices = np.append(indices, total_frames - 1)
                 video_inputs[0] = video_inputs[0][indices]
-            inputs = self.processor(text=texts, images=image_inputs, videos=video_inputs, padding=True, return_tensors="pt")
+            padding_side = "left" if self.batch_size > 1 else "right"
```
ArdalanM, line 78: These updates could be transferred to qwen2_5_vl.py

gathierry:
Hi @ArdalanM , I think there might be a little issue with this implementation for video inputs. Videos frames are sampled by `process_vision_info()` but in `self.processor()` there will be another frame sampling, resulting in much less frames than expected. This is because processor by default take input frames as 24 fps and will try to sample to 2 fps. That means the effective frame# is only 1/12 of specified.
I tested this on VideoMME and only got 55 which is much lower than the report.

ArdalanM:
Yes. It's a mix of what you said plus the fact that VideoMME prompts differs from what Qwen3VL models expects.
I fixed both and match the official VideoMME results. Will push the fix this week.

PR #893, opened 2025-11-11
[Task] Add STARE Task

pangyyyyy:
Add STARE task from https://github.com/STARE-bench/STARE/

### DIFF
--- a/lmms_eval/tasks/stare/utils.py
+++ b/lmms_eval/tasks/stare/utils.py
@@ -1,0 +1,328 @@
+import logging
+import os
+import re
+from collections import defaultdict
+from pathlib import Path
+
+import yaml
+from latex2sympy2 import latex2sympy
+from sympy import simplify
+from word2number import w2n
+
+from lmms_eval.llm_judge import get_server
+from lmms_eval.llm_judge.protocol import ServerConfig
+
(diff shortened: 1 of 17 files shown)


lmms_eval/tasks/stare/utils.py
```diff
@@ -0,0 +1,328 @@
...
+
+    # Append the selected strategy instruction
+    if strategy == "CoT":
+        query = f"{filled_prompt}\n{stare_config['Strategy_Instruction']['CoT']}"
+    else:
+        query = f"{filled_prompt}\n{stare_config['Strategy_Instruction']['Directly']}"
+
+    # Build the result dictionary
+    res_dict = {"query": query, "gt_content": answer, **sample}
+    return res_dict
```
kcz358, line 76: Would recommend to use doc_to_messages so that no need to put the placeholder for chat model and the images can be correctly interleaved

kcz358 (commented):
Hi, thank you for the contribution. I think most of the code looks nice. The build query should be fine to work for simple model as it is hard to come up with a unified solution with it. I would recommend you to use `doc_to_messages` api also so that the chat model can build the interleave format better. Feel free to refer to mmmu for more info. Thanks!

PR #894, opened 2025-11-11
[Task] Add SpatialViz Task

pangyyyyy:
Add SpatialViz Task from https://github.com/wangst0181/Spatial-Visualization-Benchmark

### DIFF
--- a/lmms_eval/tasks/spatialviz/utils.py
+++ b/lmms_eval/tasks/spatialviz/utils.py
@@ -1,0 +1,145 @@
+import os
+import re
+from collections import defaultdict
+from pathlib import Path
+
+import yaml
+from huggingface_hub import snapshot_download
+from PIL import Image
+
+with open(Path(__file__).parent / "_default_template_yaml", "r") as f:
+    raw_data = f.readlines()
+    safe_data = []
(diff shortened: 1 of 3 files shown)


lmms_eval/tasks/spatialviz/utils.py
```diff
@@ -0,0 +1,145 @@
...
+        # remove function definition since yaml load cannot handle it
+        if "!function" not in line:
+            safe_data.append(line)
+    config = yaml.safe_load("".join(safe_data))
+
+cache_dir = snapshot_download(
+    repo_id=config["dataset_path"],
+    repo_type="dataset",
+    local_dir_use_symlinks=False,
+)
```
kcz358, line 23: May I ask does this require the user to unzip the zip file by themselves? I saw that on the hub the image is in zip and video in dataset kwargs is set to false. I think the best practice for this kind of dataset is to push a convert dataset to the hub. Not sure if the cache dir under current settings works automatically. Thanks!

Luodian (approved):
LGTM. SpatialViz task implementation. Minor: consider using eval_logger instead of print in aggregate function.

PR #923, opened 2025-12-05
Add GEdit-Bench and ImgEdit-Bench for Bagel Model

KemingWu:
Add GEdit-Bench and  ImgEdit-Bench Eval Task and Modify Bagel model for inference
GEdit-Bench Results:
image

ImgEdit-Bench Results:
image

### DIFF
--- a/lmms_eval/models/simple/bagel.py
+++ b/lmms_eval/models/simple/bagel.py
@@ -234,10 +234,18 @@ class Bagel(lmms):
         else:
             self.response_persistent_folder = response_persistent_folder
 
-        if output_image_dir is None:
-            self.output_image_dir = os.path.join(self.response_persistent_folder, "bagel_generated_images")
-        else:
+        # Check for task-specific output directory from environment variables
+        # Priority: output_image_dir param > IMGEDIT_OUTPUT_DIR > GEDIT_BENCH_OUTPUT_DIR > default
+        if output_image_dir is not None:
             self.output_image_dir = output_image_dir
+        elif os.getenv("IMGEDIT_OUTPUT_DIR"):
+            self.output_image_dir = os.getenv("IMGEDIT_OUTPUT_DIR")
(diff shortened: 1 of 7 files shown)


lmms_eval/api/task.py
```diff
@@ -1057,13 +1057,21 @@ def concat_tar_parts(tar_parts, output_tar):
                 **dataset_kwargs if dataset_kwargs is not None else {},
             )
 
+        # Ensure dataset is a DatasetDict so downstream logic that expects multiple splits works.
+        if not isinstance(self.dataset, datasets.DatasetDict):
+            split_name = self.config.test_split or self.config.validation_split or self.config.training_split or "train"
+            self.dataset = datasets.DatasetDict({split_name: self.dataset})
+
```
kcz358, line 1064: This is a too ground breaking changes. Might break too much

lmms_eval/api/task.py
```diff
@@ -1057,13 +1057,21 @@ def concat_tar_parts(tar_parts, output_tar):
...
             for split in self.dataset:
                 if split in [self.config.training_split, self.config.validation_split, self.config.test_split, self.config.fewshot_split]:
                     self.dataset[split] = self.config.process_docs(self.dataset[split])
 
         # copy dataset, remove image features
-        self.dataset_no_image = self.dataset.copy()
+        try:
+            self.dataset_no_image = self.dataset.copy()
+        except AttributeError:
+            self.dataset_no_image = datasets.DatasetDict({k: v for k, v in self.dataset.items()})
```
kcz358, line 1074: Might be breaking changes also. Consider not to use this.

lmms_eval/models/simple/bagel.py
kcz358, line 1: To be honest, edit and gen doesn't seems should be fixed. It simply depends on the input.

lmms_eval/tasks/gedit_bench/secret.env
kcz358, line 1: No need to put the entire source code into here.

PR #961, opened 2025-12-30
fix: improve spatialviz utils quality

Luodian:
## Summary

Follow-up improvements to the SpatialViz task (PR #894):

- **Fix wrong exception type**: `FileExistsError` → `FileNotFoundError` (file doesn't exist, not file exists)
- **Use loguru logger**: Replace `print()` with `eval_logger` for consistent logging across the codebase
- **Add type hints**: All functions now have proper type annotations
- **Fix hidden bug**: Missing comma in `final_answer_patterns` list (`"答案" "correct path"` was concatenated as single string)
- **Remove dead code**: Redundant `image_path = image_path` assignment
- **Fix potential bug**: Initialize `op` variable to prevent `UnboundLocalError`
- **Code style**: Break long prompt string for 88-char line limit compliance


Luodian:
Retriggering CI

PR #985, opened 2026-01-12
fix: qwen2.5 video frames issue

oscarqjh:

- [ ] A detailed description

```sh
pip install pre-commit
pre-commit install
pre-commit run --all-files
```

### Ask for review
Once you feel comfortable for your PR, feel free to @ one of the contributors to review

### DIFF
--- a/lmms_eval/models/chat/qwen2_5_vl.py
+++ b/lmms_eval/models/chat/qwen2_5_vl.py
@@ -63,8 +63,6 @@ class Qwen2_5_VL(Qwen2_5_VLSimple):
             }
             if self.fps is not None:
                 video_kwargs["fps"] = self.fps
-            else:
-                video_kwargs["nframes"] = self.max_num_frames
             batched_messages = [chat_message.to_hf_messages(video_kwargs=video_kwargs) for chat_message in chat_messages]
             texts = self.processor.apply_chat_template(batched_messages, tokenize=False, add_generation_prompt=True)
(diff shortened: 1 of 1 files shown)


oscarqjh:
image

This is an attempt to fix the above issue. When I set "max_num_frames"=32 and when the benchmark contains video with less than 32 frames, it causes an error for qwenvl_2.5_chat model. (this issue does not exists for qwenvl_2.5_simple model)

oscarqjh:
@kcz358

kcz358:
If remove the nframes, then basically we are eval it with fps=2? According to 

https://github.com/QwenLM/Qwen3-VL/blob/e5c7e5c26af6a8bd65aec9388f3642cf6ea9d75c/qwen-vl-utils/src/qwen_vl_utils/vision_process.py#L167-L181

but not actually with 32 frames

Luodian:
Thanks for identifying this issue! I've created #987 with an improved fix that:

1. **Preserves `nframes` parameter** - so default behavior remains frame-count-based sampling (not FPS-based)
2. **Adds bounds check** - only resamples when `total_frames > max_num_frames`
3. **Keeps `np.unique()`** - ensures no duplicate indices

The key insight from @kcz358's comment was correct - removing `nframes` would change the default to fps=2 sampling, which alters evaluation semantics.

Closing this in favor of #987. Thanks again for the report!

Luodian:
Superseded by #987

PR #987, opened 2026-01-12
fix: properly handle qwen2.5 video frames edge case

Luodian:
## Summary

Properly fix the qwen2.5 video frames issue reported in #985.

**Problem**: When `max_num_frames` > actual video frames, `np.linspace()` creates duplicate indices causing indexing errors.

**Root cause analysis**: The original fix in #985 removed `nframes` parameter, which would change the default sampling behavior from frame-count-based to FPS-based (fps=2), altering evaluation semantics.

**This fix**:
- Keeps `nframes` parameter for proper frame-count-based sampling
- Adds bounds check: only resample if `total_frames > max_num_frames`
- Uses `np.unique()` to ensure no duplicate indices
- When video has fewer frames than `max_num_frames`, uses all frames as-is

Closes #985 

Thanks @oscarqjh for identifying the issue!

### DIFF
--- a/lmms_eval/models/chat/qwen2_5_vl.py
+++ b/lmms_eval/models/chat/qwen2_5_vl.py
@@ -19,7 +19,9 @@ from lmms_eval.protocol import ChatMessages
 try:
     from qwen_vl_utils import process_vision_info
 except ImportError:
-    eval_logger.warning("Failed to import qwen_vl_utils; Please install it via `pip install qwen-vl-utils`")
+    eval_logger.warning(
+        "Failed to import qwen_vl_utils; Please install it via `pip install qwen-vl-utils`"
+    )
 
 
 @register_model("qwen2_5_vl_chat")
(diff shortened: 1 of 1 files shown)


oscarqjh:
@Luodian Thank you for the fix, but I noticed that the solution is currently incomplete.

For reference, I am running the following task (site_bench_video) which contains a video with 31 frames:
```
# tested with
# dependenies versions
# NVIDIA-SMI 550.90.07 | Driver Version: 550.90.07 | CUDA Version: 13.1
# "torch==2.7.1",
# "torchvision==0.22.1",
NUM_FRAMES=32

CUDA_VISIBLE_DEVICES=0 accelerate launch \
    --num_processes=1 \
    --num_machines=1 \
    --mixed_precision=no \
    --dynamo_backend=no \
    --main_process_port=12346 \
    -m lmms_eval \
    --model qwen2_5_vl \
    --model_args=pretrained=Qwen/Qwen2.5-VL-3B-Instruct,max_num_frames=${NUM_FRAMES},attn_implementation=flash_attention_2,interleave_visuals=False,use_custom_video_loader=True \
    --tasks site_bench_video \
    --batch_size 1 \
    --output_path ./logs/
```

This run would result in the error:
image
This issue stems from a error inside "smart_nframes" function which is from qwen_vl_utils. This means that we have to calculate the correct nframes before the "process_vision" function.

I currently patched it locally on my own branch to solve this issue:
https://github.com/oscarqjh/lmms-eval/blob/7e31af4df7f4163dace8a8577e6385b51e7ed8d9/lmms_eval/models/chat/qwen2_5_vl.py#L83C13-L127C90

@kcz358 Please take a look too and see if my understanding is correct

PR #989, opened 2026-01-12
[feat] Add CLT and clustered standard error estimation for statistical rigor

mwxely:
## Summary

This PR adds statistically rigorous standard error estimation methods to `lmms-eval`, addressing the limitations of bootstrap-only stderr calculation.

### 1. New Standard Error Methods

| Method | Description | Use Case |
|--------|-------------|----------|
| `stderr_clt` | CLT-based SE: `std / sqrt(n)` | Fast, always available |
| `stderr_clustered` | Clustered SE for correlated questions | When questions share context (e.g., same video/image) |

### 2. New `cluster_key` Configuration

Tasks can specify a `cluster_key` in YAML to enable clustered SE calculation:

```yaml
# videomme.yaml
cluster_key: videoID  # e.g., VideoMME questions from same video are correlated
```

### 3. Standardized `score` Convention

Tasks returning dict from `process_results` should include a `score` field (0/1 correctness) for SE calculation:

```python
# Example: videomme/utils.py
return {
    "metric_name": {
        "score": 1.0 if correct else 0.0,
        "videoID": doc["videoID"],  # for clustered SE
        ...
    }
}
```

### 4. Files Changed

| File | Change |
|------|--------|
| `lmms_eval/api/task.py` | Add `cluster_key` to TaskConfig |
| `lmms_eval/api/metrics.py` | Add `clustered_stderr()` function |
| `lmms_eval/evaluator.py` | Call `calculate_clt_aggregate_metric()` |
| `lmms_eval/evaluator_utils.py` | Implement CLT/clustered SE calculation |
| `lmms_eval/utils.py` | Add `Stderr_CLT` and `Stderr_Clustered` columns |
| `lmms_eval/tasks/videomme/utils.py` | Add `score` and `videoID` to output |
| `lmms_eval/tasks/videomme/videomme.yaml` | Add `cluster_key: videoID` |

## Test Results

Full VideoMME evaluation (2700 questions) with Qwen2.5-VL-7B-Instruct:

ec8bdb872ca92d8a8473c0aa9777c4f7

- `Stderr_Clustered (0.0097) > Stderr_CLT (0.0094)` indicates correlation between questions from the same video
- Both values are much smaller than smoke test (n=10) due to larger sample size

## Backward Compatibility

- Existing tasks without `cluster_key` show `N/A` for clustered stderr
- Tasks not returning `score` in dict show `N/A` for both CLT stderrs
- Output format is additive (new columns, no breaking changes)

## How to Enable for Other Tasks

1. Add `cluster_key: <field_name>` to task YAML
2. Include `score` (0/1) and the cluster field in `process_results` output dict

### DIFF
--- a/lmms_eval/utils.py
+++ b/lmms_eval/utils.py
@@ -588,7 +588,10 @@ def make_table(result_dict, column: str = "results", sort_results: bool = False)
 
             # Helper to format stderr value
             def fmt_se(se_val):
-                if se_val is None or se_val == "N/A" or se_val == []:
+                if se_val is None or se_val == "N/A":
+                    return "N/A"
+                # Handle empty list/array safely (numpy array comparison is ambiguous)
+                if hasattr(se_val, "__len__") and len(se_val) == 0:
                     return "N/A"
                 try:
                     return "%.4f" % se_val
@@ -602,7 +605,9 @@ def make_table(result_dict, column: str = "results", sort_results: bool = False)
             # Clustered stderr
             se_clustered = fmt_se(dic.get(m + "_stderr_clustered," + f))
 
-            if v != []:
+            # Check if v is not empty (handle numpy array safely)
+            is_empty = hasattr(v, "__len__") and not isinstance(v, str) and len(v) == 0
+            if not is_empty:
                 values.append([k, version, f, n, m, hib, v, "±", se, se_clt, se_clustered])
             # k = ""
             # version = ""


lmms_eval/api/metrics.py
```diff
@@ -644,3 +644,43 @@ def aggregate_subtask_metrics(metrics, sizes, weight_by_size=True):
...
+
+    # Calculate cluster means
+    cluster_means = [sum(s) / len(s) for s in cluster_scores.values()]
+
+    n_clusters = len(cluster_means)
+    if n_clusters < 2:
+        return float("nan")
+
+    # Calculate SE of cluster means
+    import numpy as np
```
kcz358: This one seems redundant

lmms_eval/api/metrics.py
```diff
@@ -644,3 +644,43 @@ def aggregate_subtask_metrics(metrics, sizes, weight_by_size=True):
...
+
+    return np.std(cluster_means, ddof=1) / np.sqrt(n_clusters)
```
kcz358: Equation seems to be incorrect? Here calculate the std of cluster means and divide by 1/n_clusters. From the paper, should first calculate the variance for total scores + the variance in-clustered then sqrt.

lmms_eval/evaluator_utils.py
```diff
@@ -122,6 +125,35 @@ def calculate_aggregate_metric(self, bootstrap_iters=100000) -> None:
...
+            # Convention: dict items should have "score" (0/1) and optionally the field specified by cluster_key
+            numeric_items = []
+            cluster_ids = []
+            for x in items:
+                if isinstance(x, (int, float)):
+                    numeric_items.append(x)
+                    cluster_ids.append(None)
+                elif isinstance(x, dict) and "score" in x:
+                    numeric_items.append(x["score"])
+                    cluster_ids.append(x.get(cluster_key) if cluster_key else None)
```
kcz358, line 147: Maybe we should add a score key later in the task config as well.

mwxely:
I have addressed all the feedback:

1. **Redundant import** - Removed `import numpy as np` inside `clustered_stderr()`, now using module-level `import collections` and `math.sqrt()`.

2. **Formula correction** - Rewrote `clustered_stderr()` to implement Equation 4 from the paper (arxiv:2411.00640):
   ```
   SE_clustered = sqrt(SE_CLT² + (1/n²) * Σ_c Σ_i Σ_{j≠i} (s_ic - s̄)(s_jc - s̄))
   ```
   The cross-term now correctly accounts for within-cluster correlations.

3. **Configurable score_key** - Added `score_key` field to `TaskConfig` (default: `"score"`), allowing tasks to customize which dict field contains 0/1 scores.

Commits:
- `bc07ff2` fix(metrics): correct clustered_stderr formula per Eq.4 from paper
- `db7adf7` feat(config): add score_key option for configurable score field

lmms_eval/api/metrics.py
```diff
@@ -651,36 +652,49 @@ def clustered_stderr(scores: List[float], cluster_ids: List[Any]) -> float:
...
+
+    # Calculate within-cluster cross-terms: sum_c sum_i sum_{j!=i} (s_ic - s_bar)(s_jc - s_bar)
+    cross_term = 0.0
+    for cid, cluster_scores in cluster_to_scores.items():
+        # For each cluster, compute sum of (s_i - s_bar)(s_j - s_bar) for i != j
+        deviations = [s - s_bar for s in cluster_scores]
+        cluster_sum = sum(deviations)
+        # sum_{i!=j} d_i * d_j = (sum d_i)^2 - sum(d_i^2)
+        sum_of_squares = sum(d * d for d in deviations)
+        cross_term += cluster_sum * cluster_sum - sum_of_squares
```
kcz358, line 695: A small question here. I feel this is simply calculating the sum of (s_i - s_bar) and then ^2 of it? Still not equivalent to the equation.

I feel an efficient implementation here is such that for each cluster. 

Calculate A = deviations * deviation^T and the cross term should add sum(A) - trace(A)

lmms_eval/evaluator_utils.py
```diff
@@ -129,20 +129,21 @@ def calculate_clt_aggregate_metric(self) -> None:
...
                 continue
             # Extract scores and cluster_ids from items
-            # Convention: dict items should have "score" (0/1) and optionally the field specified by cluster_key
+            # Convention: dict items should have the score_key field (0/1) and optionally the field specified by cluster_key
             numeric_items = []
             cluster_ids = []
             for x in items:
                 if isinstance(x, (int, float)):
                     numeric_items.append(x)
                     cluster_ids.append(None)
```
kcz358, line 144: Could remove this if, otherwise seems might add possible unrelated info?

lmms_eval/api/metrics.py
mwxely: ### TL;DR

The current implementation is actually **mathematically equivalent** to the matrix formulation you suggested, but **more efficient**.

---

## Mathematical Equivalence Proof

The target formula from [Evan Miller's paper](https://arxiv.org/abs/2411.00640) Equation (4):

$$SE_{clustered} = \sqrt{SE_{CLT}^2 + \frac{1}{n^2} \sum_c \sum_i \sum_{j \neq i} (s_{i,c} - \bar{s})(s_{j,c} - \bar{s})}$$

For each cluster, we need to compute:
$$\sum_i \sum_{j \neq i} d_i d_j \quad \text{where } d_i = s_i - \bar{s}$$

### Current Implementation (Algebraic Trick)

```python
deviations = [s - s_bar for s in cluster_scores]  # d_i
cluster_sum = sum(deviations)                      # Σd_i
sum_of_squares = sum(d * d for d in deviations)   # Σd_i²
cross_term += cluster_sum * cluster_sum - sum_of_squares
```

**Mathematical identity used:**

$$\left(\sum_i d_i\right)^2 = \sum_i \sum_j d_i d_j = \sum_i d_i^2 + \sum_{i \neq j} d_i d_j$$

Therefore:

$$\sum_{i \neq j} d_i d_j = \left(\sum_i d_i\right)^2 - \sum_i d_i^2$$

### Matrix Implementation (Your Suggestion)

```python
deviations = np.array(cluster_scores) - s_bar
A = np.outer(deviations, deviations)  # A_ij = d_i * d_j
cross_term += np.sum(A) - np.trace(A)
```

Where:
- `sum(A)` = $\sum_i \sum_j d_i d_j$ = $(\sum_i d_i)^2$ = `cluster_sum²`
- `trace(A)` = $\sum_i d_i^2$ = `sum_of_squares`
- `sum(A) - trace(A)` = $\sum_{i \neq j} d_i d_j$

**Both implementations compute exactly the same value.**

---

## Time Complexity Proof

### Current Implementation (Algebraic)

For a cluster with $m$ samples:
- Computing `deviations`: $O(m)$
- Computing `cluster_sum = sum(deviations)`: $O(m)$
- Computing `sum_of_squares = sum(d*d)`: $O(m)$
- Computing `cluster_sum² - sum_of_squares`: $O(1)$

**Total per cluster: $O(m)$**

**Total for all clusters: $O(n)$** where $n$ is total samples

Space complexity: $O(m)$ for storing deviations list (can be reduced to $O(1)$ with streaming)

### Matrix Implementation

For a cluster with $m$ samples:
- Computing `deviations`: $O(m)$
- Computing `A = np.outer(deviations, deviations)`: $O(m^2)$
- Computing `np.sum(A)`: $O(m^2)$
- Computing `np.trace(A)`: $O(m)$

**Total per cluster: $O(m^2)$**

**Total for all clusters: $O(\sum_c m_c^2)$** which is $O(n \cdot \bar{m})$ where $\bar{m}$ is average cluster size

Space complexity: $O(m^2)$ for storing matrix A

### Summary

| Implementation | Time Complexity | Space Complexity |
|---------------|-----------------|------------------|
| **Algebraic (current)** | $O(n)$ | $O(m)$ |
| Matrix | $O(n \cdot \bar{m})$ | $O(m^2)$ |

The algebraic implementation is **asymptotically faster** and **more memory efficient**.

---

## Experimental Verification

### Test Configuration
- **Task**: videomme
- **Samples**: 300
- **Clusters**: 100 (3 questions per video)
- **Model**: Qwen2.5-VL-7B-Instruct

### Results

```
=================================================================
IMPLEMENTATION COMPARISON (LIMIT=300, 100 clusters)
=================================================================
CLT stderr (naive):            0.025152082937712
Clustered (algebraic/current): 0.025039921055704
Clustered (matrix/kcz358):     0.025039921055704
-----------------------------------------------------------------
Difference:                    3.47e-18
=================================================================
```

### Performance Benchmark (300 clusters × 3 items = 900 samples)

```
Current impl (algebraic): 0.886 ms
Matrix impl (outer):      2.863 ms
Speedup:                  3.23x
```

---

## Conclusion

The current algebraic implementation is:
1. **Mathematically equivalent** to the matrix formulation (verified: diff = 3.47e-18)
2. **More efficient**: O(n) vs O(n·m̄), ~3x faster in practice
3. **More memory efficient**: O(m) vs O(m²)

lmms_eval/evaluator_utils.py
mwxely: ### TL;DR

This branch is intentionally kept for **backward compatibility**. Here is the design philosophy:

1. **Some tasks return raw numeric scores** (int/float) instead of dict. Without this branch, those tasks would silently skip CLT stderr calculation.

2. **When items are numeric**, we set `cluster_ids.append(None)`. This means `valid_clusters` will be empty (line 152), and `stderr_clustered` will correctly return `"N/A"` (line 155-156). **No "unrelated info" is added to the clustered calculation.**

3. **CLT stderr still works** for tasks without cluster configuration - it just reports naive standard error.

The current implementation handles both cases cleanly:

| Input Type | cluster_key configured | CLT stderr | Clustered stderr |
|------------|----------------------|------------|------------------|
| Dict with score | Yes | ✅ Computed | ✅ Computed |
| Dict with score | No | ✅ Computed | N/A |
| Raw numeric | N/A | ✅ Computed | N/A |

This ensures maximum compatibility across different task configurations while producing correct results.

---

## Summary

| Comment | Action Required | Reason |
|---------|-----------------|--------|
| Comment 1 (formula) | No code change | Implementation is correct and efficient |
| Comment 2 (if branch) | No code change | Design is intentional for backward compatibility |

kcz358 (approved):

mwxely:
**Error:**
```
ValueError: The truth value of an empty array is ambiguous. Use array.size > 0 to check that an array is not empty.
```

**Fix (commit `ef7a943`):**
```python
# New code - safe for both list and numpy array
if se_val is None or se_val == "N/A":
    return "N/A"
if hasattr(se_val, "__len__") and len(se_val) == 0:
    return "N/A"
```

This successfully fixes the errors during `cicd` test:  

1cf107e54e4b78ca3130d96157b1a1ef

PR #991, opened 2026-01-12
Add CaptionQA benchmark task

bronyayang:
CaptionQA evaluates how well image captions preserve information for downstream QA tasks. The benchmark uses:
- A vision-language model to generate captions
- Qwen2.5-72B-Instruct as a judge to answer questions based on captions
- Scoring based on correctness with partial credit for 'cannot answer'

Features:
- Works out of the box: no environment variables needed
- Automatic deferred judge evaluation (runs after caption model unloads)
- Supports 4 domains: natural, document, ecommerce, embodiedai
- Uses vLLM for efficient 72B judge inference with TP=2

Usage:
    python -m lmms_eval \
        --model qwen2_5_vl \
        --model_args pretrained=Qwen/Qwen2.5-VL-3B-Instruct \
        --tasks captionqa \
        --batch_size 1 \
        --output_path ./logs/captionqa_results

Paper: https://arxiv.org/abs/2511.21025
Dataset: https://huggingface.co/datasets/Borise/CaptionQA

@Luodian  @kcz358  @pufanyi

### DIFF
--- a/lmms_eval/tasks/captionqa/utils.py
+++ b/lmms_eval/tasks/captionqa/utils.py
@@ -56,12 +56,8 @@ _global_shuffle_initialized = False
 # The 72B judge model requires at least 2 GPUs with tensor parallelism
 JUDGE_BACKEND = os.getenv("CAPTIONQA_JUDGE_BACKEND", "vllm")  # "vllm" or "api"
 JUDGE_MODEL = os.getenv("CAPTIONQA_JUDGE_MODEL", "Qwen/Qwen2.5-72B-Instruct")
-JUDGE_TP_SIZE = int(
-    os.getenv("CAPTIONQA_JUDGE_TP_SIZE", "2")
-)  # Default 2 for 72B model
-JUDGE_GPU_MEMORY_UTILIZATION = float(
-    os.getenv("CAPTIONQA_JUDGE_GPU_UTIL", "0.8")
-)  # GPU memory for judge
+JUDGE_TP_SIZE = int(os.getenv("CAPTIONQA_JUDGE_TP_SIZE", "2"))  # Default 2 for 72B model
+JUDGE_GPU_MEMORY_UTILIZATION = float(os.getenv("CAPTIONQA_JUDGE_GPU_UTIL", "0.8"))  # GPU memory for judge
 
 # System prompt for QA evaluation (matching original)
 QA_SYSTEM_PROMPT = "You are given a caption describing an image, and a question about the image. Answer with a SINGLE LETTER (A, B, C, ...), no explanation."
@@ -120,9 +116,7 @@ def _initialize_global_shuffle_cache():
     try:
         from datasets import load_dataset
 
-        eval_logger.info(
-            "Initializing global shuffle cache from 'all' split (matching original RNG order)..."
-        )
+        eval_logger.info("Initializing global shuffle cache from 'all' split (matching original RNG order)...")
         dataset = load_dataset("Borise/CaptionQA", split="all")
 
         # Use the exact same RNG setup as the original
@@ -184,9 +178,7 @@ def _initialize_global_shuffle_cache():
                 question_count += 1
 
         _global_shuffle_initialized = True
-        eval_logger.info(
-            f"Global shuffle cache initialized: {question_count} questions cached"
-        )
+        eval_logger.info(f"Global shuffle cache initialized: {question_count} questions cached")
 
     except Exception as e:
         eval_logger.warning(f"Failed to initialize global shuffle cache: {e}")
@@ -223,15 +215,10 @@ def get_shuffle_permutation(image_id: str, q_idx: int, n_choices: int) -> List[i
         if len(cached_perm) == n_choices:
             return cached_perm.copy()
         else:
-            eval_logger.warning(
-                f"Cached permutation length mismatch for ({image_id}, {q_idx}): "
-                f"cached={len(cached_perm)}, expected={n_choices}. Using fallback."
-            )
+            eval_logger.warning(f"Cached permutation length mismatch for ({image_id}, {q_idx}): " f"cached={len(cached_perm)}, expected={n_choices}. Using fallback.")
 
     # Fallback to hash-based seed if not in cache
-    eval_logger.debug(
-        f"Question ({image_id}, {q_idx}) not in shuffle cache, using hash-based seed"
-    )
+    eval_logger.debug(f"Question ({image_id}, {q_idx}) not in shuffle cache, using hash-based seed")
     question_seed = hash((image_id, q_idx, SHUFFLE_SEED)) % (2**32)
     rng = random.Random(question_seed)
     perm = list(range(n_choices))
@@ -249,9 +236,7 @@ def _get_vllm_judge():
         import vllm
         from transformers import AutoTokenizer
 
-        eval_logger.info(
-            f"Initializing vLLM judge with model: {JUDGE_MODEL}, tp_size: {JUDGE_TP_SIZE}"
-        )
+        eval_logger.info(f"Initializing vLLM judge with model: {JUDGE_MODEL}, tp_size: {JUDGE_TP_SIZE}")
 
         tokenizer = AutoTokenizer.from_pretrained(JUDGE_MODEL, trust_remote_code=True)
         llm = vllm.LLM(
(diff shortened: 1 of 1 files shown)


lmms_eval/tasks/captionqa/run_captionqa.py
kcz358, line 1: This seems to be a debug code for running the task?

lmms_eval/tasks/captionqa/run_judge.py
kcz358, line 1: This also seems to be a debug code for running evaluate and process result. Maybe can remove both

lmms_eval/tasks/captionqa/eval_precomputed.py
kcz358, line 1: This part seems to be the alignment code? Maybe can also remove if not needed in the eval logic.

lmms_eval/tasks/captionqa/_judge_subprocess.py
kcz358, line 1: This one is also the debug code?

lmms_eval/tasks/captionqa/utils.py
```diff
@@ -0,0 +1,999 @@
...
+# The original uses a single sequential RNG (seed=0) that advances through ALL questions
+# in the "all" split order. To reproduce exact shuffles, we precompute all permutations.
+_global_shuffle_cache: Dict[tuple, List[int]] = {}
+_global_shuffle_initialized = False
+
+# Two-phase evaluation mode
+# CAPTIONQA_SKIP_JUDGE: Skip judge entirely (for manual two-phase workflow)
+# CAPTIONQA_DEFER_JUDGE: Defer judge to aggregation phase (runs after lm.clean() frees GPU)
+SKIP_JUDGE = os.getenv("CAPTIONQA_SKIP_JUDGE", "0").lower() in ("1", "true", "yes")
+DEFER_JUDGE = os.getenv("CAPTIONQA_DEFER_JUDGE", "1").lower() in ("1", "true", "yes")  # Default ON
```
kcz358: Recommend to use a one phase evaluation so that one evaluating your task can do it in one go. So can simplify your logic here if possible

bronyayang:
Thanks for checking! I just realized that a lot of debug code is not
removed and comments are not updated. Let me clean the code and commit
again.

lmms_eval/tasks/captionqa/utils.py
```diff
@@ -0,0 +1,999 @@
...
+    _judge_client = {"server": server, "config": server_config, "type": "api"}
+    return _judge_client
+
+
+def _get_judge():
+    """Get the appropriate judge client based on configuration."""
+    if JUDGE_BACKEND == "vllm":
+        return _get_vllm_judge()
+    else:
+        return _get_api_judge()
```
kcz358, line 295: For launching a local vlm/llm as judge, we recommend two ways

1. You could do it with lmms_eval internal launcher 

https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/a3fad2fd2607d89c28bba1a56be4ec2abd9cd567/lmms_eval/evaluator.py#L521-L526

but we only write one simple sglang launcher though. This one is used just in case you have to squeeze in your eval server and eval model in one node so we clean up the memory and then launch it with subprocess with you.

2. If your cluster have enough gpu to setup a separate eval server, can setup a separate server using sglang/vllm and then just use openai client to call the eval. The port or base url etc. can be set using your unique env var. (I use this way a lot though in my daily usage, examples in https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/main/lmms_eval/tasks/_task_utils/reasoning_utils.py where I usually setup a different eval server)

To conclude my suggestion is that we do suggest that you can use a unified client to keep the code clean. Thanks if this is possible!

kcz358 (commented):
Hi, thank you for the contribution. Most of the code and logic looks fine to me. There are just some debugging codes and logic I observe in your code seems are uncleaned for a PR. Would recommend to keep the pipeline clean by removing these codes. Also, feel free to see if the eval launcher in lmms_eval or separate launching eval server on different nodes would be useful to you. This allows you no need to setup a separate vllm LLM in your eval logic. Thanks!

mwxely:
Hi, there. Just a friendly reminder - the lint check is failing due to formatting issues in `lmms_eval/models/chat/qwen2_5_vl.py`. Running `pre-commit run --all-files` locally and then committing the changes should fix it.

bronyayang:
Hi lmms-eval team,
Thanks for reviewing! I've cleaned up the code based on feedback:

Changes made:
1. Removed all debugging code and extra scripts (run_captionqa.py, run_judge.py, eval_precomputed.py)
2. Simplified to single-stage evaluation
3. Cleaned up outdated comments and dead code
4. Ran pre-commit run --all-files - all checks pass (black, isort)

Usage is now simple, as an example:
python -m lmms_eval \    --model qwen2_5_vl \    --model_args pretrained=Qwen/Qwen2.5-VL-3B-Instruct \    --tasks captionqa \    --batch_size 1 \    --output_path ./logs/captionqa_results

Requirements:
- 2+ GPUs with ~80GB VRAM each (for the 72B judge model)
- vLLM installed
The judge evaluation runs automatically during the aggregation phase after the caption model unloads, so everything happens in one command.

Files (8 total):
utils.py - core evaluation logic
captionqa.yaml + 4 domain configs
_default_template_yaml - task template
_judge_subprocess.py - subprocess helper for GPU isolation

Let me know if you need any other changes! :) @kcz358

lmms_eval/tasks/captionqa/utils.py
```diff
@@ -0,0 +1,852 @@
...
```
kcz358, line 295: This part hasn't been resolved? Would not recommend to init an actual vllm LLM object. Only recommend use the oai judge api. If you have to squeeze the server launching in the GPUs, please consider using the eval launcher as I mentioned in previous comments.

lmms_eval/tasks/captionqa/utils.py
```diff
@@ -0,0 +1,852 @@
...
+                _global_shuffle_cache[(image_id, q_idx)] = perm.copy()
+                question_count += 1
+
+        _global_shuffle_initialized = True
+        eval_logger.info(f"Global shuffle cache initialized: {question_count} questions cached")
+
+    except Exception as e:
+        eval_logger.warning(f"Failed to initialize global shuffle cache: {e}")
+        eval_logger.warning("Falling back to per-question hash-based seeds")
+        _global_shuffle_initialized = True  # Prevent repeated attempts
```
kcz358, line 186: Recommend to use process docs instead of setting a glob here. If you evaluate using vllm etc, sometimes different process might not setting the correct glob. Check 

https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/4814efac2eaf68f9f2e02181d6f7530f95713ef5/lmms_eval/tasks/charxiv/utils.py#L38-L57

for an example

lmms_eval/tasks/captionqa/utils.py
```diff
@@ -0,0 +1,852 @@
...
+        # Cleanup temp files
+        if os.path.exists(input_path):
+            os.unlink(input_path)
+        if os.path.exists(output_path):
+            os.unlink(output_path)
+
+    _deferred_eval_done = True
+    eval_logger.info(f"\nDeferred evaluation complete. Evaluated {len(_evaluated_results_cache)} questions.\n")
+
+    return _evaluated_results_cache
```
kcz358, line 783: Running eval in subprocess seems to be quite unsafe and not really recommended. Is it possible to not splitting here? Seems like the subprocess is simply calling the evaluation single question, wondering if can call this inside the process result.

kcz358 (changes requested):
Hi there are a few suggestions for refactor, otherwise seems the code logic is not capsulated very well.

1. Consider using only the openai client. Setting up a single vllm LLM is not recommended. If you hope to squeeze your eval server along with the evaluation process, please consider using the eval launcher in lmms-eval and specify in your task readme.
2. Consider not calling the evaluate single question in subprocess. Saving pickles and then calling the judge api in another file seems to be very dirty logic. I believe this could be called in the process results? So you don't have to use placeholder for your return score etc.
3. Use process result to process dataset if you want to get perm etc. You can have access to the doc during process result.

Please let me know if these changes are reasonable and pratical to you. Thanks

PR #992, opened 2026-01-13
fix: qwen2.5vl nframes bug

oscarqjh:
Please see my comments here @Luodian @kcz358 
https://github.com/EvolvingLMMs-Lab/lmms-eval/pull/987#issuecomment-3742679213

Currently there are alot of debugging code, I'm opening a pull request to bring your attention to this issue. Will clean up the code after we come to a conclusion. Thank you.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ uv.lock
 workspace/*
 .claude/*
 remote_code/*
+.opencode
--- a/lmms_eval/api/metrics.py
+++ b/lmms_eval/api/metrics.py
@@ -1,10 +1,11 @@
 # the code is adapted from https://github.com/EleutherAI/lm-evaluation-harness
+import collections
 import math
 import random
 import re
 import string
 from collections.abc import Iterable
-from typing import List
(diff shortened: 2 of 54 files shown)


kcz358:
Hi, I think your understanding is correct. I do find out this issue a while ago but I realize this is the smart n frames problem in the qwen vl utils so I switch to use fps + max frames since I feel this should be a better way to sample frames dynamically. So yeah I think I am not surprise that #987 is not fixing this issue and thus I would still recommend to use fps + max frames to better sample videos.

oscarqjh:
> Hi, I think your understanding is correct. I do find out this issue a while ago but I realize this is the smart n frames problem in the qwen vl utils so I switch to use fps + max frames since I feel this should be a better way to sample frames dynamically. So yeah I think I am not surprise that #987 is not fixing this issue and thus I would still recommend to use fps + max frames to better sample videos.

Okay thank you for the clarification! I'll use fps + max frames moving forward.

Luodian (approved):

PR #996, opened 2026-01-15
[Task] Added SiteBench multi-image variant and bug fix

oscarqjh:
1. Added option to input video as multi image to align with models like SenseNova-SI qwen series that are trained on multi image input. See https://github.com/EvolvingLMMs-Lab/EASI/issues/20.

sample run result:
image

2. Fixed the issue in official code where post-prompt and pre-prompt are not extracted properly from lmms_eval_specific_kwargs

### DIFF
--- a/lmms_eval/tasks/sitebench/utils.py
+++ b/lmms_eval/tasks/sitebench/utils.py
@@ -70,11 +70,7 @@ def parse_multi_choice_response(response, all_choices):
     # Look for choices with periods, e.g., A:, B:, C:.
     if len(candidates) == 0:
         for choice in all_choices:
-            if (
-                f"{choice}:" in response
-                or f":{choice}" in response
-                or f": {choice}" in response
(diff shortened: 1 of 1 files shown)


oscarqjh:
@PeterWangyi @kcz358

oscarqjh:
b792f6c75036c971f995aba13fe90012629f3a09 added option to use interleave_visual - this is required to align evaluation result with VLMEvalKit

PeterWangyi:
It seems that in the Internvl series models, we can only call the doc_to_visual function, but not the doc_to_message function.
This will cause the text-image interleave to fail.

oscarqjh:
> It seems that in the Internvl series models, we can only call the doc_to_visual function, but not the doc_to_message function. This will cause the text-image interleave to fail.

I think this should be a model problem, not a task problem - will address it in a different issue here #998

Issue #998, opened 2026-01-16
[Bug] InternVL3: Duplicate <image>  tokens when input already contains image placeholders


oscarqjh:
## Description
When evaluating InternVL3 models on tasks where the input text already contains `<image>` tokens (e.g., interleaved image-text tasks), the model wrapper unconditionally prepends additional `<image>` tokens, resulting in duplicate placeholders and degraded performance.

## Current Behavior
In `internvl3.py`, the `generate_until` method always prepends image tokens regardless of whether they already exist in the input:

This causes issues when:

The task's doc_to_text already includes `<image>` placeholders for interleaved image-text formats
Multi-image tasks where image positions matter (e.g., "Compare image 1 `<image>` with image 2 `<image>`")
Example:

If the input context is:
```
What is the difference between <image> and <image>?
```

The current code transforms it to:
```
<image> <image>
What is the difference between <image> and <image>?
```

This results in 4 image tokens but only 2 images, causing misalignment and incorrect model outputs.

Example ran on SiteBench:

<img width="811" height="222" alt="Image" src="https://github.com/user-attachments/assets/922b3123-b0ca-4ee2-a105-d8dec40c96f3" />

## Expected Behavior
The model should:

Check if `<image>` tokens already exist in the input context
Only prepend image tokens if they are not already present
Validate that the number of `<image>` tokens matches the number of images

## Impact
This bug affects evaluation accuracy on:

- Interleaved image-text benchmarks
- Multi-image comparison tasks
- Tasks with custom image placeholder positions

PR #999, opened 2026-01-16
[Bug] internvl3 duplicate <image> token issue

oscarqjh:
## Description

This pull request directly attempts to address the issue at #998

## Summary
This PR fixes a bug in the InternVL3 model integration where the <image> token was being inserted twice in certain prompting scenarios. This redundancy could lead to incorrect model behavior or input format errors.

## Details of Changes
Modified the logic in models/simple/internvl3.py to prevent appending an extra <image> token if one is already present in the prompt template.

Ensured that the input passed to the tokenizer contains exactly one image placeholder per image.

## Test Plan
- [x] Verified with: Ran InternVL3 on SiteBenchImage.

image

- [x] Observation: Checked the logs/debug output to confirm that the prompt now contains the correct number of <image> token as expected.

image

### DIFF
--- a/lmms_eval/models/simple/internvl3.py
+++ b/lmms_eval/models/simple/internvl3.py
@@ -106,18 +106,10 @@ def dynamic_preprocess(
     orig_width, orig_height = image.size
     aspect_ratio = orig_width / orig_height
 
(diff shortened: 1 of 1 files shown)


oscarqjh:
@PeterWangyi @kcz358

PR #1012, opened 2026-01-20
[Model] Bagel UMM

oscarqjh:
## Add BagelUMM Model Integration
This PR adds support for the Bagel multimodal model to lmms-eval, enabling evaluation of Bagel's image understanding, generation, and editing capabilities.

#### New Files
`lmms_eval/models/simple/bagel_umm.py`
 - Complete Bagel model integration

#### Modified Files
`lmms_eval/models/init.py`
 - Register bagel_umm in AVAILABLE_SIMPLE_MODELS
pyproject.toml

#### Supported Modes

| Mode              | Description                                                                 |
|-------------------|-----------------------------------------------------------------------------|
| understanding     | Visual question answering – the model analyzes images and answers questions about them. Default mode. |
| think_understanding | Same as `understanding` but with chain-of-thought reasoning before answering. |
| generate          | Text-to-image generation.                                                   |
| think_generate    | Text-to-image generation with thinking process.                             |
| editImage         | Image editing based on text instructions.                                   |
| think_editImage   | Image editing with thinking process.                                        |

#### Key Implementation Details
- Interleaved Input Support: Properly handles `<image>` tokens for multi-image interleaved prompts
- Mode-specific Parameters: Each mode has optimized default hyperparameters (CFG scales, timesteps, etc.) aligned with VLMEvalKit - see here https://github.com/EvolvingLMMs-Lab/VLMEvalKit/blob/cf9ee4628fde0b5396ecbe1a3383bfdb3a7b63b6/vlmeval/vlm/bagel_umm.py#L22C1-L76C2

### Tested With
- NVIDIA Driver 550.90.07 / CUDA 13.1
- PyTorch 2.7.1 / torchvision 0.22.1
- flash-attn 2.8.3
- Python 3.11

### Example Results
image


oscarqjh:
@PeterWangyi @kcz358

PeterWangyi:
Is it possible to integrate this with bagel.py? Since it's essentially the same model architecture, it would also help reduce confusion.

oscarqjh:
> Is it possible to integrate this with bagel.py? Since it's essentially the same model architecture, it would also help reduce confusion.

I believe that it can directly replace bagel.py. May I seek some assistance to verify this?

kcz358:
Yeah, I did a quick glance on this and seems it mostly are just the bagel.py. Also, if you guys are using bagel from lmms-engine, can also try using bagel_lmms_engine model in the chat folder.

oscarqjh:
> Yeah, I did a quick glance on this and seems it mostly are just the bagel.py. Also, if you guys are using bagel from lmms-engine, can also try using bagel_lmms_engine model in the chat folder.

Oh, sorry I would like to clarify if the current bagel.py implementation supports multimodal (image + text ->  text) generation?
When I attempted to run benchmarks like VSI-Bench and SiteBench using the existing `bagel.py`, it seemed limited to image generation (or lacked the correct text-generation pipeline), which is why I wrote a new `bagel_umm.py`.

Could you advise on the best path forward? Should I replace the current bagel.py entirely, or would it be better to merge my changes into the existing file?

kcz358:
I would actually suggest to replace the bagel.py since I believe the inferencer kind of can do both the understanding and the generation part? Possibly is just the original bagel.py does not pass in the correct parameters or arguments. Not sure if the bagel_lmms_engine did it in correct way.

PR #1016, opened 2026-01-22
add structeditbench task

KemingWu:
image


lmms_eval/tasks/structeditbench/utils.py
```diff
@@ -0,0 +1,502 @@
...
+                    break
+                except Exception:
+                    continue
+            if isinstance(v, str) and os.path.exists(v):
+                try:
+                    edited_image_pil = Image.open(v).convert("RGB")
+                    edited_image_path = v
+                    break
+                except Exception:
+                    continue
```
kcz358, line 417: Just found out that this part can be much much more precise though. Way too much guarding in my opinion

Issue #845, opened 2025-10-01
Request for Meta-Llama-3-8B-Instruct when evaluating LLaVA-Video-7B with conv_template=qwen_1_5


Ethan16162:
I am using the following script to evaluate **llava-video-7B**:

```bash
accelerate launch --num_processes=4 \
-m lmms_eval \
--model llava_vid \
--model_args pretrained=/data/gys/models/LLaVA-Video-7B-Qwen2,conv_template=qwen_1_5,max_frames_num=64,mm_spatial_pool_mode=average \
--tasks mme \
--batch_size 1 \
--log_samples \
--log_samples_suffix llava_vid \
--output_path ./logs/
```
But I encountered the following error:
```
'(MaxRetryError("HTTPSConnectionPool(host='huggingface.co', port=443): Max retries exceeded with url: /meta-llama/Meta-Llama-3-8B-Instruct/resolve/main/tokenizer_config.json (Caused by ConnectTimeoutError(<urllib3.connection.HTTPSConnection object at 0x7f32141fb7f0>, 'Connection to huggingface.co timed out. (connect timeout=10)'))"), '(Request ID: 9d6f8820-9bec-4a91-ae0c-b189630c8e26)')' thrown while requesting HEAD https://huggingface.co/meta-llama/Meta-Llama-3-8B-Instruct/resolve/main/tokenizer_config.json
```
However, I have already specified ` conv_template=qwen_1_5 `. Why is `Meta-Llama-3-8B-Instruct` still appearing here?

rookiexiong7:
same issue

Nudavia:
same issue

Nudavia:
I solve this promble by add HF_ENDPOINT and HUGGINGFACE_TOKEN in my scripts, but it always stuck(or too slow), the log is：

`Rank 0:  Loaded LLaVA model: /node2/disk2/model_weight/llava-hf/llava-onevision-qwen2-7b-ov-hf
You are using a model of type llava_onevision to instantiate a model of type llava_qwen. This is not supported for all configurations of models and can yield errors.
You are using a model of type llava_onevision to instantiate a model of type llava_qwen. This is not supported for all configurations of models and can yield errors.
You are using a model of type llava_onevision to instantiate a model of type llava_qwen. This is not supported for all configurations of models and can yield errors.
Rank 0:  Overwriting config with {'mm_spatial_pool_stride': 2, 'mm_spatial_pool_mode': 'bilinear'}
[...]

PR #1024, opened 2026-01-22
fix: replace hardcoded .cuda() with .to(self._device) for multi-GPU support

### LINKED ISSUE #952: Qwen2.5-VL Expect all tensors on the same devices
Dear author,
      when I evaluate Qwen2.5-VL-3B on only one GPU and set num_processes=1, the error occurred!
"Error during evaluation: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument index in method wrapper_CUDA__index_select). Please set `--verbosity=DEBUG` to get more information."

I've searched issues for this , but all are about multiple GPUS, so could you help me solve this. ‼️I've tried "device_map=auto",but it not works
My .sh file is as follows 

export HF_HOME="~/.cache/huggingface"
export CUDA_VISIBLE_DEVICES=0

accelerate launch --num_processes=1 --main_process_port=12344 -m lmms_eval \
    --model qwen2_5_vl \
    --model_args=pretrained=./datasets/Qwen2.5-VL-3B-Instruct,max_pixels=12845056,attn_implementation=flash_attention_2,interleave_visuals=False,device_map=auto \
    --tasks mathvision_test \
    --device cuda:0 \
    --output_path ./logs/ \
    --batch_size 4

Luodian:
## Summary

Replaces hardcoded `.cuda()` calls with `.to(self._device)` to support proper device placement when using multi-GPU setups or non-default CUDA devices.

## Problem

Many model implementations use hardcoded `.cuda()` calls which always place tensors on `cuda:0`, regardless of the `device` parameter passed to the model. This causes issues when:
- Running on specific GPUs (e.g., `--device cuda:1`)
- Using multi-GPU distributed evaluation
- The default GPU is occupied or unavailable

## Changes

Fixed device placement in 9 major model implementations:
- `llava_onevision.py` - 3 occurrences
- `llava_vid.py` - 4 occurrences
- `vila.py` - 6 occurrences
- `longva.py` - 1 occurrence
- `internvl.py` - 4 occurrences
- `internvl2.py` - 2 occurrences
- `internvideo2.py` - 4 occurrences (including helper function fix)
- `internvideo2_5.py` - 3 occurrences + model loading order fix
- `auroracap.py` - 3 occurrences

All `.cuda()` calls replaced with `.to(self._device)` to respect the device configuration.

## Related Issues

Fixes #952, #845

## Testing

- Verified all modified files have valid Python syntax
- No new lint errors introduced (existing pre-existing warnings remain)
- Device placement now respects `self._device` which is properly set based on accelerator configuration

## Note

24 `.cuda()` calls remain in less commonly used models (cambrian, egogpt, videochat, vita, etc.) - these can be addressed in a follow-up PR if needed.

### DIFF
--- a/lmms_eval/models/simple/auroracap.py
+++ b/lmms_eval/models/simple/auroracap.py
@@ -37,9 +37,7 @@ try:
     )
     from lmms_eval.models.aurora_xtuner.utils import PROMPT_TEMPLATE
 except ImportError:
-    eval_logger.error(
-        "AuroraCap is not installed. Please install AuroraCap to use this model by `git clone https://github.com/rese1f/aurora.git` and link `src/xtuner/xtuner` to `lmms_eval/models/aurora_xtuner`"
-    )
+    eval_logger.error("AuroraCap is not installed. Please install AuroraCap to use this model by `git clone https://github.com/rese1f/aurora.git` and link `src/xtuner/xtuner` to `lmms_eval/models/aurora_xtuner`")
 import warnings
 
 warnings.filterwarnings("ignore")
@@ -92,11 +90,7 @@ class AuroraCap(lmms):
         else:
             self._device = device
 
-        pretrained_pth = (
-            snapshot_download(repo_id=pretrained)
-            if not osp.isdir(pretrained)
-            else pretrained
-        )
+        pretrained_pth = snapshot_download(repo_id=pretrained) if not osp.isdir(pretrained) else pretrained
         pretrained_llm = pretrained_pth
         pretrained_vit = osp.join(pretrained_pth, "visual_encoder")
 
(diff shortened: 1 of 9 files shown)


Luodian:
## Review Notes

This is exactly the kind of fix that saves hours of debugging. When someone runs `--device cuda:1` and tensors still land on `cuda:0`, the error messages are cryptic.

**Scope:** 9 model files, 30+ `.cuda()` → `.to(self._device)` replacements. The changes are mechanical but the impact is significant.

**What's good:**

1. **Fixes real issues** - Addresses #952 and #845 directly
2. **Consistent pattern** - All replacements follow `x.to(self._device)` 
3. **Minimal diff** - Just the device placement changes plus some formatting cleanup

**One observation:**

The PR mentions 24 `.cuda()` calls remain in less common models. That's fine for a follow-up, but worth tracking. Partial fixes can be confusing - users might assume device selection works everywhere after this PR.

**Formatting changes:**

Mixed in with the device fixes are line-length reformats (splitting long lines). This is fine but makes the diff noisier. Consider separating pure-format commits from behavior changes in the future.

**Overall:** Important fix. Multi-GPU setups are increasingly common and this unblocks a whole class of users. ✓

kcz358 (approved):

PR #1026, opened 2026-01-22
fix: add resource cleanup in video loaders to prevent memory leaks

Luodian:
## Summary

Adds proper resource cleanup in video loading utilities to prevent memory leaks and file handle exhaustion when processing large numbers of videos.

## Problem

When running evaluations on video datasets, the video loaders were not properly releasing resources:
- `load_video_decord()`: VideoReader objects were not explicitly deleted
- `read_video_pyav()`: av containers were not explicitly closed

This leads to memory buildup and potential "too many open files" errors during long-running evaluations.

## Changes

1. **load_video_decord()**: Added `del vr` after extracting frames to release VideoReader
2. **read_video_pyav()**: Wrapped in try/finally to ensure `container.close()` is always called
3. Replaced bare `except:` with `except Exception:` for better error handling practices

## Testing

- Verified syntax is valid
- No functional changes to frame extraction logic
- Resources are now properly released after use

### DIFF
--- a/lmms_eval/models/model_utils/load_video.py
+++ b/lmms_eval/models/model_utils/load_video.py
@@ -16,9 +16,7 @@ def load_video_decord(video_path, max_frames_num):
     else:
         vr = VideoReader(video_path[0], ctx=cpu(0))
     total_frame_num = len(vr)
-    uniform_sampled_frames = np.linspace(
-        0, total_frame_num - 1, max_frames_num, dtype=int
-    )
+    uniform_sampled_frames = np.linspace(0, total_frame_num - 1, max_frames_num, dtype=int)
     frame_idx = uniform_sampled_frames.tolist()
     spare_frames = vr.get_batch(frame_idx).asnumpy()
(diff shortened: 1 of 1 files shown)


Luodian:
## Review Notes

This fixes a real problem. Video evaluations that process thousands of files will eventually hit file handle limits without proper cleanup.

**What's good:**

1. `del vr` after `get_batch()` - Releases the VideoReader immediately rather than waiting for GC
2. `try/finally` with `container.close()` - Guarantees cleanup even if exceptions occur
3. Bare `except:` → `except Exception:` - Good practice. Bare except catches SystemExit and KeyboardInterrupt too.

**One thought:**

The `del vr` approach works, but a context manager would be cleaner:

```python
with VideoReader(video_path, ctx=cpu(0)) as vr:
    # use vr
```

If decord's VideoReader doesn't support `with`, then `del` is the right call. Worth checking.

**Overall:** Solid defensive fix. Will prevent mysterious "too many open files" errors in production.

kcz358 (approved):

PR #1038, opened 2026-01-24
feat: add reasoning task versions for multiple benchmarks

kcz358:
## Summary
Add reasoning task versions for multiple benchmarks with `<thinking>` and `<answer>` tags.

## Changes
- **ai2d**: Add reasoning task with thinking/answer tags
- **realworldqa**: Add reasoning task with thinking/answer tags
- **mmbench (en)**: Add dev and test reasoning tasks with thinking/answer tags
- **phyx**: Add reasoning tasks for mc, oe, mini_mc, and mini_oe variants with thinking/answer tags
- **ocrbench_v2**: Add reasoning task with thinking/answer tags, reusing original process_results
- **olympiadbench**: Add reasoning tasks for math and physics in English and Chinese with thinking/answer tags

## Implementation Details
All reasoning tasks follow a consistent pattern:
- Use system prompt requiring `<thinking>...</thinking>` and `<answer>...</answer>` tags
- Use `compute_score` from `reasoning_utils` for scoring (except ocrbench_v2 which reuses original)
- Support `acc_score` and `format_score` metrics
- Use interleaved message format for image/text content

### DIFF
--- a/lmms_eval/tasks/olympiadbench_mimo/reasoning/olympiadbench_math_en_reasoning.yaml
+++ b/lmms_eval/tasks/olympiadbench_mimo/reasoning/olympiadbench_math_en_reasoning.yaml
@@ -3,7 +3,7 @@
 dataset_path: lscpku/OlympiadBench-official
 dataset_kwargs:
   token: True
-  dataset_name: maths_en_all
+dataset_name: maths_en_all
 task: olympiadbench_math_en_reasoning
 test_split: test
 output_type: generate_until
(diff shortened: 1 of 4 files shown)


kcz358:
image

Samples for using the reasoning variant with OpenMMReasoner. The doc to target changes is for a fix to correctly show the target for the ocrbench otherwise in the jsonl, the target would just show the "answer" str since a mismatch on type

Luodian:
Got it. The `doc_to_target` fix makes sense - you need the actual target value in the logs, not the string "answer".

The screenshot shows the reasoning format working with OpenMMReasoner. Looks good.

One thing though - you didn't address the tag mismatch question. Your system prompts use `<think>...</think>` but the PR description says `<thinking>...</thinking>`. Which one does `reasoning_utils.compute_score()` actually parse? Just want to make sure they match.

Otherwise this is ready to merge.

kcz358:
`<think>` tag is the actual tag, the PR description is a mismatch. GLM seems to have some issue for outputting the think token

PR #1039, opened 2026-01-25
[Model]: add InternVL-HF model support

CodeTeamster:
## Summary
Add InternVL-HF model support for multi-modal evaluation, specifically for InternVL3-HF and InternVL3.5-HF. The adaptation leverages the implementation of the InternVL model using the HuggingFace transformers library. Supports both image and video inputs.
 
## Changes
* **InternVL3-HF**: Add support for the latest official hf-format model such as https://huggingface.co/OpenGVLab/InternVL3-8B-hf
* **InternVL3.5-HF**: Add support for the latest official hf-format model such as https://huggingface.co/OpenGVLab/InternVL3_5-8B-HF
* **InternVL-HF Base Class**: Implement version-agnostic base class `InternVLHf` for all hf-format InternVL variants
* **Multi-modal Pipeline**: Maintain compatibility with existing image and video processing pipelines 

## Implementation Details
The InternVL-HF integration follows a consistent implementation pattern:
 
* Uses the official HuggingFace transformers implementation
* Designed with forward compatibility for potential future releases of InternVL1.0-2.5 HF weights
* Maintains compatibility with existing image and video processing pipelines
* Ready to support official InternVL1.0-2.5 HF weights if released in the future on the huggingface.

## Ask for review

### DIFF
--- a/examples/models/internvl_hf.sh
+++ b/examples/models/internvl_hf.sh
@@ -1,0 +1,20 @@
+export HF_HOME="~/.cache/huggingface"
+export CUDA_VISIBLE_DEVICES=0,1,2,3
+
+# bash ./internvl_hf.sh "mme" "/path/to/InternVL3_5-8B-HF"
+
+NUM_DEVICES=$(echo $CUDA_VISIBLE_DEVICES | awk -F',' '{print NF}')
+TASK=$1
+CKPT_PATH=$2
+echo $TASK
+TASK_SUFFIX="${TASK//,/_}"
+echo $TASK_SUFFIX
+
+accelerate launch --num_processes $NUM_DEVICES --main_process_port 12380 -m lmms_eval \
+    --model internvl_hf \
+    --model_args pretrained=$CKPT_PATH \
+    --tasks $TASK \
+    --batch_size 1 \
+    --log_samples \
+    --log_samples_suffix $TASK_SUFFIX \
+    --output_path ./logs/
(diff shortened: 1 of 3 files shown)


CodeTeamster:
> Solid implementation. The code follows the established patterns from other HF models.
> 
> **Quick observations:**
> 
> **Batch size check** Line 86 asserts batch_size must be 1. Is this a temporary limitation or architectural? If it's just for initial release, worth noting in a TODO.
> 
> **Device map logic** The device selection logic (lines 94-103) handles single GPU, multi-GPU, and auto cases. Clean, but three different paths for device assignment could get confusing. Works fine though.
> 
> **Video handling** Line 276: you set `videos = None` if empty list. Good defensive programming - prevents passing empty tensors.
> 
> **Generation kwargs** Lines 289-296 set defaults for max_new_tokens, temperature, etc. These match other models in the codebase, which is what you want for consistency.
> 
> **Missing: loglikelihood** You raise NotImplementedError for loglikelihood. Fair enough - most vision models don't need it. But some benchmarks (MMBench) use it for multiple choice. Worth documenting which tasks work/don't work with this model.
> 
> Overall: clean, follows conventions, should work fine. Merge when ready.

Thank you so much for taking the time to review this PR and for your thoughtful feedback. I really appreciate your positive comments on the implementation.

Here are my responses to your specific points:

**Batch size check**
The image processor of InternVL model is implemented based on GotOcr2ImageProcessor. This processor adopts the dynamic tiling logic, typically assumes that the number and size of patches for each sample vary and are processed dynamically on a per-sample basis. As a result, the engineering implementation enforces a batch size of 1 (as seen in the assert statement), making it not directly feasible to use batch sizes greater than 1 by default.

**Device map logic**
Thanks for noting this. I agree that having three distinct paths could be simplified. Therefore, I've reduced the number of branches to two by performing a forced commit.

**Missing: loglikelihood**
Log-likelihood computation for InternVLHf will be implemented in the next commit, so I think it is appropriate to add a TODO note here.

Issue #1037, opened 2026-01-24
How do I run a particular sample


chiungyit:
I want to reproduce the issue on a erring sample without running the whole dataset. Is there a way to do so? 

I'm using an older version.
```
uv run "$PYTHON" -m lmms_eval  -h
usage: __main__.py [-h] [--config CONFIG] [--model MODEL] [--tasks TASKS] [--model_args MODEL_ARGS]
                   [--launcher_args LAUNCHER_ARGS] [--num_fewshot NUM_FEWSHOT] [--batch_size auto|auto:N|N]
                   [--max_batch_size N] [--device DEVICE] [--output_path = [dir/file.jsonl] [DIR]] [--limit LIMIT]
                   [--use_cache DIR] [--cache_requests {true,refresh,delete}] [--check_integrity] [--write_out]
                   [--log_samples] [--wandb_log_samples] [--log_samples_suffix LOG_SAMPLES_SUFFIX]
                   [--system_instruction SYSTEM_INSTRUCTION] [--apply_chat_template] [--fewshot_as_multiturn]
                   [--show_config] [--include_path INCLUDE_PATH] [--gen_kwargs GEN_KWARGS] [--verbosity VERBOSITY]
                   [--wandb_args WANDB_ARGS] [--timezone TIMEZONE] [--hf_hub_log_args HF_HUB_LOG_ARGS]
                   [--predict_only] [--seed SEED] [--trust_remote_code] [--process_with_media] [--force_simple]

options:
  -h, --help            show this help message and exit
  --config CONFIG       Path to a yaml file specifying all eval arguments, will ignore cli arguments if specified
  --model MODEL         Name of model e.g. `hf`
  --tasks TASKS         To get full list of tasks, use the command lmms-eval --tasks list
  --model_args MODEL_ARGS
                        String arguments for model, e.g. `pretrained=EleutherAI/pythia-160m,dtype=float32`
  --launcher_args LAUNCHER_ARGS
                        String arguments for launcher for local llm as judge, e.g. `tp=8`, if None then no launcher will be used.
  --num_fewshot NUM_FEWSHOT
                        Number of examples in few-shot context
  --batch_size auto|auto:N|N, -b auto|auto:N|N
                        Acceptable values are 'auto', 'auto:N' or N, where N is an integer. Default 1.
  --max_batch_size N    Maximal batch size to try with --batch_size auto.
[...]

PR #1042, opened 2026-01-27
Add --offset option

chiungyit:
https://github.com/EvolvingLMMs-Lab/lmms-eval/issues/1037

- [ ] A detailed description

```sh
pip install pre-commit
pre-commit install
pre-commit run --all-files
```

### Ask for review
Once you feel comfortable for your PR, feel free to @ one of the contributors to review


Luodian:
Nice implementation. The offset logic is clean - you're correctly slicing the dataset after loading.

A few things to verify:
1. Does offset work with caching? Line 789 in evaluator.py shows you're applying it to filtered requests. Make sure cached requests get offset applied too.
2. Edge case: what happens if offset > dataset size? Should fail gracefully with a clear error.
3. The CICD test would be good - try running one of the existing test tasks with `--offset 5 --limit 10` to verify it processes samples 5-14.

The maintainer's suggestion about CICD is spot on. Add a quick test case to verify offset+limit interaction.

Otherwise looks good. This solves a real pain point for debugging specific samples.

chiungyit:
@Luodian Good point! 
1. The use case this PR targets is to run particular samples with --limit and --offset. This is often executed after the full run of benchmark is completed and the faulty samples are identified from cache or result jsonl. That said, by the time, the cache of full benchmark run has been created.  So I feel referencing to modifying existing entries in the cache doesn't make sense. It's cleanest if the users save the new cache and result somewhere else, but they don't, append at the end.

Besides, from my previous [experiment](https://github.com/EvolvingLMMs-Lab/lmms-eval/issues/1037#issuecomment-3802339023), once --limit is specified, the cache are not recognized. The new entries are appended at the end. Would it be cleaner to follow the existing behavior, i.e. Add the new caching entries at the end, disregard the existing entries.

How do you think?

2, 3, Agree, will do

kcz358:
I feel caching can be ignored under this case since this is mainly for debugging and it is okay to rerun the samples (which sometimes we want to do so)

PR #1049, opened 2026-01-28
fix incomplete configs of mmmu_val*

Salieri0515:
add `dataset_name: Accounting` and comments to `mmmu_val*.yaml`

mmmu_val*.yaml lacks `dataset_name`, leading to `Value Error: Config name is missing`

@Luodian

### DIFF
--- a/lmms_eval/tasks/mmmu/mmmu_val.yaml
+++ b/lmms_eval/tasks/mmmu/mmmu_val.yaml
@@ -1,4 +1,6 @@
 dataset_path: lmms-lab/MMMU
+# Change dataset_name to evaluate on different subjects.
+dataset_name: Accounting
 task: "mmmu_val"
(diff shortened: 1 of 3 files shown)


Luodian (commented):
This fixes the ValueError but raises a question.

You're setting `dataset_name: Accounting` as the default. But MMMU has multiple subjects. Why Accounting specifically?

Looking at the validation split, users probably want to run all subjects, not just Accounting. Should the default be to run everything? Or should the config be more explicit about this being a single-subject example?

If Accounting is just a placeholder, the comment should make that clearer. Something like:

```yaml
# REQUIRED: Set this to the subject you want to evaluate
# Available subjects: Accounting, Agriculture, Architecture, Art, ...
dataset_name: Accounting
```

What do you think?

PR #1050, opened 2026-01-28
[Task] add PAIBench-U

Jialuo-Li:
Hello.

This PR adds a new benchmark PAIBench-U, which aims to evaluate MLLM's capability of video understanding in Physical AI domain.

Arxiv: https://arxiv.org/abs/2512.01989

Data: https://huggingface.co/datasets/Jialuo21/PAIBench-U

Thank you!

### DIFF
--- a/lmms_eval/tasks/paibench_u/utils.py
+++ b/lmms_eval/tasks/paibench_u/utils.py
@@ -1,6 +1,7 @@
 import os
 import re
 from pathlib import Path
+
 import yaml
 from loguru import logger as eval_logger
 
@@ -20,13 +21,12 @@ cache_dir_test = os.path.join(base_cache_dir, cache_name_test)
 
 
 def paibench_u_doc_to_visual(doc):
-    video_path = os.path.join(cache_dir_test, 'videos', doc["video_path"])
+    video_path = os.path.join(cache_dir_test, "videos", doc["video_path"])
     if os.path.exists(video_path):
         return [video_path]
     else:
         eval_logger.warning(f"Video path: {video_path} does not exist, skipping sample")
         return []
-
 
 
 def paibench_u_doc_to_text(doc, lmms_eval_specific_kwargs=None):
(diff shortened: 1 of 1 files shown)


Luodian (commented):
Clean implementation. The task follows the established patterns.

**Observations:**

**Video path handling**
Line 37 calls sys.exit() if the video doesn't exist. That crashes the entire evaluation run instead of skipping one sample or logging an error. Consider returning None or raising a more graceful exception that the eval framework can catch.

**Random fallback in parser**
Lines 78-83: if parsing fails and no letter is found, you return 'Unknown'. But the comment says 'assign a random choice'. Which is it? The code doesn't match the comment. If you want random fallback, you need random.choice(['A', 'B', 'C', 'D', 'E']).

**Parsing robustness**
The parse_multi_choice_response function strips common answer prefixes. Good. But it only searches for single letters A-E. What if the model outputs '(A)' or '[A]' or 'A)'? Worth testing with a few models to see if you need more robust parsing.

**Metric naming**
Line 20: metric name is 'paibench_u_percetion_score' - typo? Should be 'perception'.

**Missing README**
No README.md documenting the task. Users won't know what PAIBench-U tests or how to run it. Add a brief README with paper link, dataset description, and example usage.

**Unused imports**
Line 5: 'from random import random' - never used. Line 3: 'import json' - never used. Clean these up.

Otherwise solid. The aggregation logic looks correct - overall, category, and subcategory breakdowns make sense for this kind of benchmark.

Jialuo-Li:
Thanks for the detailed review. I’ve gone ahead and addressed those points to tighten up the implementation. Regarding the README, we actually already have the task documentation, paper links hosted in our official repository, so we’re keeping the documentation centralized there for now.

kcz358:
LGTM! Can you run a pre-commit to pass the lint test. Thanks! Then I will merge this PR.

Jialuo-Li:
Fixed!

PR #1064, opened 2026-02-03
[Task] Add CoreCognition bench

Irisicy4:
Description:
- support additional single-image-at-start-of-prompt task **_CoreCognition_** 
- with options to only run subset since the benchmark is 3-staged
- with options to supply api and run a hybrid answer matching method (llm match the failed-to-be-template-matched answers) so it reproduces the original paper's setup
- see task and paper details in task README

### Ask for review

Thank you!

### DIFF
--- a/lmms_eval/tasks/corecognition/utils.py
+++ b/lmms_eval/tasks/corecognition/utils.py
@@ -19,11 +19,13 @@ eval_logger = logging.getLogger("lmms-eval")
 _default_template_path = Path(__file__).parent / "_default_template_yaml"
 _corecognition_config_path = Path(__file__).parent / "corecognition.yaml"
 
+
 def _load_yaml_stripped(path: Path) -> dict:
     with open(path, "r") as f:
         raw_data = f.readlines()
     safe_data = [line for line in raw_data if "!function" not in line]
     return yaml.safe_load("".join(safe_data)) or {}
+
 
 _corecognition_config = _load_yaml_stripped(_default_template_path)
 _corecognition_config.update(_load_yaml_stripped(_corecognition_config_path))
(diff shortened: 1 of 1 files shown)


lmms_eval/tasks/corecognition/utils.py
```diff
@@ -0,0 +1,284 @@
...
+from typing import Any, Dict, Optional
+
+import yaml
+
+eval_logger = logging.getLogger("lmms-eval")
+
+# Load task config for use_lmms_judge (strip !function lines so yaml.safe_load works)
+_corecognition_config_path = Path(__file__).parent / "corecognition.yaml"
+_corecognition_config = None
+if _corecognition_config_path.exists():
```
kcz358: Can assume always exists. No need for extra safeguarding in my opinion.

lmms_eval/tasks/corecognition/corecognition.yaml
kcz358, line 1: Can create a _default_template_yaml for all the yaml file and include it to reduce repeated code.

lmms_eval/tasks/corecognition/utils.py
Irisicy4: edited in last commit

lmms_eval/tasks/corecognition/corecognition.yaml
Irisicy4: edited in last commit

Irisicy4:
Thanks for reviewing @Luodian . One further question --
while in this PR I committed the frame-combined version with only a single image like attached, our paper's full dataset contains interleaved multi-image and video.
For a potential iteration, would LMMS be able to support such complexity in one yaml if you can share some examples? Or I split full dataset by modality like multi-image one task + video one task?
Alternatively, if only the single-image version is encouraged for model compatibility free feel to advise so.

Thank you!

image

PR #1069, opened 2026-02-07
[Fix] Add dynamic max_num calculation to InternVL3 to align with VLMEvalKit

oscarqjh:
## Summary

This PR aligns the InternVL3 image tiling strategy with VLMEvalKit's implementation by introducing dynamic `max_num` calculation. This ensures consistent evaluation results between lmms-eval and VLMEvalKit when benchmarking InternVL3 models.

### New Parameter: `total_max_num`

Added a new model argument `total_max_num` (default: 64) that sets the total budget for image tiles across all images in a single inference.

```python
# Model initialization
model_args = "pretrained=OpenGVLab/InternVL3-8B,max_num=12,total_max_num=64"
```

### Dynamic `max_num` Calculation

Instead of using a fixed `max_num` for all images, the tile count per image is now dynamically calculated based on the number of images:

```python
# For images:
dynamic_max_num = max(1, min(self.max_num, self.total_max_num // image_num))

# For video (each frame treated as an image):
dynamic_max_num = max(1, min(self.max_num, self.total_max_num // self.num_frame))
```

**Behavior:**
- Single image: Uses full `max_num` (e.g., 12 tiles)
- Multiple images: Tiles are distributed to stay within `total_max_num` budget
  - 2 images → up to 12 tiles each (64 // 2 = 32, capped by max_num=12)
  - 8 images → up to 8 tiles each (64 // 8 = 8)
  - 16 images → up to 4 tiles each (64 // 16 = 4)

This prevents OOM issues with multi-image inputs while maximizing resolution for single-image benchmarks.

## Why This Matters

VLMEvalKit uses this dynamic tiling strategy, which means:
1. **Reproducibility**: Results from lmms-eval now match VLMEvalKit evaluations
2. **Fair comparison**: Benchmark scores are directly comparable across frameworks
3. **Memory efficiency**: Multi-image/video inputs won't exceed memory limits

## Files Changed

- [lmms_eval/models/simple/internvl3.py](lmms_eval/models/simple/internvl3.py)
  - Added `total_max_num` parameter to `__init__`
  - Implemented dynamic `max_num` calculation for image modality
  - Implemented dynamic `max_num` calculation for video modality

## Backward Compatibility

- Default values (`max_num=12`, `total_max_num=64`) match VLMEvalKit defaults
- Existing configurations without `total_max_num` will work unchanged
- For single-image benchmarks, behavior is identical to before (12 tiles max)

## Related

- VLMEvalKit InternVL implementation: https://github.com/open-compass/VLMEvalKit/blob/main/vlmeval/vlm/internvl/internvl_chat.py

### DIFF
--- a/lmms_eval/tasks/mmsearch_plus/decrypt_utils.py
+++ b/lmms_eval/tasks/mmsearch_plus/decrypt_utils.py
@@ -8,11 +8,11 @@ from typing import Any, Dict
 def derive_key(password: str, length: int) -> bytes:
     """
     Derive encryption key from password using SHA-256.
-    
+
     Args:
         password: Password/canary string
         length: Desired key length
-        
+
     Returns:
         Derived key of specified length
     """
@@ -25,11 +25,11 @@ def derive_key(password: str, length: int) -> bytes:
 def decrypt_text(ciphertext_b64: str, password: str) -> str:
     """
     Decrypt base64-encoded ciphertext using XOR cipher with derived key.
-    
+
     Args:
         ciphertext_b64: Base64-encoded encrypted string
         password: Password/canary string
-        
+
     Returns:
         Decrypted string
     """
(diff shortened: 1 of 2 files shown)


oscarqjh:
@PeterWangyi

PR #1070, opened 2026-02-08
refactor: introduce manifest-driven model registry v2

Luodian:
## Why
The current model registration path is hard to maintain:

- model definitions are duplicated across `AVAILABLE_SIMPLE_MODELS` and `AVAILABLE_CHAT_TEMPLATE_MODELS`
- resolution behavior (chat-vs-simple preference) is implemented in multiple places
- plugin loading mutates registry state at import time
- there is no load-time validation that resolved classes are valid `lmms` models with consistent `is_simple` behavior

This PR introduces a single, explicit resolution path while keeping existing user-facing behavior stable.

### 1) Manifest-driven registry (v2)
Added `ModelRegistryV2` with a minimal `ModelManifest`:

```python
@dataclass(frozen=True)
class ModelManifest:
    model_id: str
    simple_class_path: str | None = None
    chat_class_path: str | None = None
    aliases: tuple[str, ...] = ()
```

Resolution rules are centralized and explicit:
- default: prefer `chat_class_path` when available
- `force_simple=True`: use `simple_class_path` if present, otherwise fallback to chat
- aliases resolve to canonical `model_id` before class loading

### 2) Load-time class validation
`get_model_class()` now validates:
- resolved class is a subclass of `lmms`
- `is_simple` is consistent with resolved model type (`simple` vs `chat`)

This prevents silent misrouting into the wrong evaluation pipeline.

### 3) Integration points updated
- `lmms_eval/models/__init__.py`: delegates `get_model()` to registry v2
- `lmms_eval/entrypoints/http_server.py`: `/models` uses unified listing path
- `lmms_eval/tui/discovery.py`: discovery uses registry v2 manifest/list APIs
- `lmms_eval/models/model_manifests.py`: curated built-in manifests

### 4) Tests + docs
- added/updated unit tests for resolution semantics, aliases, `force_simple`, and validation checks
- updated model guide documentation for v2 registration flow

### 1) Why use v2?
Because the current dual-dict design has no single source of truth and causes duplicated, implicit resolution logic. V2 gives one manifest per model, one resolver, explicit alias handling, and load-time validation.

### 2) Why not directly replace every external contract now?
We keep compatibility for users and downstream tooling:
- `get_model(model_name, force_simple=...)` remains available
- legacy dict exports remain available
- `LMMS_EVAL_PLUGINS` path is still supported while we provide migration direction

So internals are modernized, while public entry behavior stays stable.

### 3) Why is a large replacement acceptable in this part?
This is a foundational internal boundary (`lmms_eval/models`) that all callsites depend on. Keeping two parallel resolution systems would increase long-term risk and inconsistency. A single internal cutover here reduces complexity, and safety is controlled via compatibility shims + targeted tests.

## Compatibility
- `get_model(model_name, force_simple=...)` remains compatible
- legacy dict exports remain available for compatibility
- `LMMS_EVAL_PLUGINS` path remains supported (with migration direction toward entry points)

## Validation
```bash
python -m unittest test.eval.test_model_registry_v2 -v
```

Covers:
- chat/simple preference
- alias resolution
- `force_simple` edge behavior
- non-`lmms` class rejection
- `is_simple` mismatch detection

### DIFF
--- a/lmms_eval/models/__init__.py
+++ b/lmms_eval/models/__init__.py
@@ -12,12 +12,7 @@ from lmms_eval.models.model_manifests import CORE_MODEL_MANIFESTS
 from lmms_eval.models.registry_v2 import ModelManifest, ModelRegistryV2
 
 logger.remove()
-log_format = (
-    "<green>{time:YYYY-MM-DD HH:mm:ss}</green> | "
-    "<level>{level: <8}</level> | "
-    "<cyan>{name}</cyan>:<cyan>{function}</cyan>:<cyan>{line}</cyan> - "
-    "<level>{message}</level>"
-)
+log_format = "<green>{time:YYYY-MM-DD HH:mm:ss}</green> | " "<level>{level: <8}</level> | " "<cyan>{name}</cyan>:<cyan>{function}</cyan>:<cyan>{line}</cyan> - " "<level>{message}</level>"
 logger.add(sys.stdout, level="WARNING", format=log_format)
 
 
@@ -145,16 +140,8 @@ def _build_builtin_manifests() -> list[ModelManifest]:
         manifests.append(
             ModelManifest(
                 model_id=model_id,
-                simple_class_path=(
-                    _build_class_path(model_id, "simple", simple_class)
-                    if simple_class
-                    else None
-                ),
-                chat_class_path=(
-                    _build_class_path(model_id, "chat", chat_class)
-                    if chat_class
-                    else None
-                ),
+                simple_class_path=(_build_class_path(model_id, "simple", simple_class) if simple_class else None),
+                chat_class_path=(_build_class_path(model_id, "chat", chat_class) if chat_class else None),
             ),
         )
     return manifests
@@ -166,8 +153,7 @@ def _merge_legacy_plugin_models(registry: ModelRegistryV2) -> None:
         return
 
     warnings.warn(
-        "LMMS_EVAL_PLUGINS is deprecated. Prefer Python entry-points group "
-        "'lmms_eval.models' for plugin model registration.",
+        "LMMS_EVAL_PLUGINS is deprecated. Prefer Python entry-points group " "'lmms_eval.models' for plugin model registration.",
         DeprecationWarning,
         stacklevel=2,
     )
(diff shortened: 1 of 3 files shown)


kcz358 (approved):
From my opinion, according to Occam's Razor, maybe should not create such a complicate class as ModelManifest and creating too much function as this.

If trying to rfc the model registry, I think a much more simple way is simply create a register class and then allow to pass in 2 key. For example, a simple decorator class here 

https://github.com/EvolvingLMMs-Lab/lmms-engine/blob/91558bb08453b08e0591262772f69dc8fb4ad206/src/lmms_engine/models/monkey_patch.py#L11-L28

https://github.com/EvolvingLMMs-Lab/lmms-engine/blob/91558bb08453b08e0591262772f69dc8fb4ad206/src/lmms_engine/models/qwen3/monkey_patch.py#L30

this allows a cleaner register. But I'm also okay with this seems this should also work and we just need to change the old registry to implement new models.

Luodian:
Good point - you're right that the current `ModelManifest` was heavier than necessary.

I reviewed the `lmms-engine` monkey patch pattern you linked. The decorator style is clean for in-repo registration.

In this PR, we also need to support a slightly different scope:

- unify `AVAILABLE_SIMPLE_MODELS` + `AVAILABLE_CHAT_TEMPLATE_MODELS` into one lookup
- resolve aliases (for example, `vllm_chat` -> `vllm`)
- support plugin entry points from external packages

A pure decorator alone does not cover plugin models that are not in this codebase.

I agree we should keep the manifest minimal for now. I will trim it to:

```python
@dataclass(frozen=True)
class ModelManifest:
    model_id: str
    simple_class_path: str | None = None
    chat_class_path: str | None = None
    aliases: tuple[str, ...] = ()
```

We can add fields back later only when there is a concrete consumer.

I am also planning a follow-up PR to remove the old `@register_model` decorator from all model files, since `MODEL_REGISTRY` is no longer in the resolve path.

kcz358 (commented):
I'm okay with most of the changes. Just seems the lmms_eval/models/model_manifests.py is a little bit of redundant? Hand curated a few core model manifests into a single file. Seems already got registered in the __init__.py. LGTM for the other changes, to add in new models, we still change the available chat model registry.

PR #1075, opened 2026-02-10
[Bug Fix] Cambrians model crashes on multi-image benchmarks

oscarqjh:
## Summary

  - Fix a `RuntimeError: Tensors must have same number of dimensions: got 3 and 4` crash
    when running the Cambrians model on multi-image benchmarks (e.g. MindCube)
  - The root cause was that only a single `<image>` token was inserted into the prompt
    regardless of the actual number of images, causing the downstream cambrian library
    to take its single-image code path while the image features tensor was shaped for
    multiple images
  - Now uses `len(visual_sizes)` to insert one `<image>` token per image so the cambrian
    library correctly routes to its multi-image concatenation path

  ## Test 
Verified to run on mindcube-tiny (a multi-image benchmark)
image


oscarqjh:
@PeterWangyi @kcz358

PR #1205, opened 2026-02-25
fix: replace 113 bare excepts with except Exception

haosenwang1018:
Replace 113 bare `except:` with `except Exception:` across 58 files. Bare `except:` catches `KeyboardInterrupt` and `SystemExit`, masking real errors and preventing clean process termination during long evaluation runs.


haosenwang1018:
CI failures require manual fix. Closing per policy.

PR #1208, opened 2026-02-26
feat(tasks): add reasoning collection for LLaVA-OV 1.5 RL

kcz358:
## Summary
Add reasoning tasks for multiple benchmarks with doc_to_messages for <analysis>/<answer> format and support for accuracy and format scoring metrics.

## Benchmarks Added
- MMStar
- MMBench (cn/en with dev/test splits)
- MME-RealWorld (cn/en)
- SEED-Bench
- SEED-Bench-2-Plus
- CV-Bench (2d/3d)
- OCR-Bench
- ChartQA
- InfoVQA
- CountBenchQA
- VStar-Bench
- PixMo-Count

This is an implementation of LLaVA-OV 1.5 RL reasoning collection.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -34,16 +34,154 @@ A minimal evaluation runs the entire pipeline with a single command. This exampl
 ```bash
 export OPENAI_API_KEY="your-api-key"
 
-* **[Commands Guide](commands.md)** - Learn about command line flags and options
-* **[Quick Start](quickstart.md)** - Evaluate your model in 5 minutes
-* **[Model Guide](model_guide.md)** - How to add and integrate new models
-* **[Task Guide](task_guide.md)** - Create custom evaluation tasks
-* **[Current Tasks](current_tasks.md)** - List of all supported evaluation tasks
-* **[Run Examples](run_examples.md)** - Example commands for running evaluations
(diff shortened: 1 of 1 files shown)


Luodian:
I'd like to put it target to `dev-v0d7` first and you may need to align results on this branch.

Luodian:
Since I am planning to have some designs on supporting instruct/reasoning models evaluation, I think this should mostly set on models side.

kcz358:
While I tried to use the reasoning tag in task yaml, I observe that if using the original task yaml, there might be conflict on reasoning system prompt and the post prompt (where model may answer in not aligned behavior). From the alignment perspective, I do recommend to create another task yaml to keep both of the settings so that they won't affect each other while of course we can also use the reasoning tag to filter the answer.

Luodian:
can we just add some reasoning specific field in original yaml?

kcz358:
Commit message for [4cada1c](https://github.com/EvolvingLMMs-Lab/lmms-eval/pull/1208/commits/4cada1c44b0f26b7179fafecc2e043bfc9220cca) is wrong, it is because there is a duplicated reasoning tag args so I remove it.

PR #1218, opened 2026-03-02
fix: handle empty OpenAI-compatible text responses

andyluo7:
## Summary

Harden the OpenAI-compatible evaluation path so `None` or structured `message.content` values do not break lmms-eval postprocessing.

## Changes

- normalize OpenAI-compatible `message.content` so `None` becomes an empty string and list-style structured content is flattened into plain text
- guard `unwrap_generation_output()` against raw `None` and `GenerationResult(text=None)`
- guard `ConfigurableTask.process_results()` so generate-until results are coerced to strings before `.strip()`

## Why

Some OpenAI-compatible backends can legally return `message.content = None` or structured content blocks. Previously those values propagated into task postprocessing and caused crashes like:

`AttributeError: 'NoneType' object has no attribute 'strip'`

## Validation

- verified the patched files parse cleanly
- confirmed OCRBench evaluation completes successfully after this fix

### DIFF
--- a/lmms_eval/api/instance.py
+++ b/lmms_eval/api/instance.py
@@ -48,11 +48,9 @@ def unwrap_generation_output(output: Any) -> Tuple[str, Optional[TokenCounts]]:
     maximum backward compatibility.
     """
     if isinstance(output, GenerationResult):
-        return "" if output.text is None else str(output.text), output.token_counts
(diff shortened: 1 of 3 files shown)


Luodian:
I made the change to let the modification minimal, since that's enough to guard the output is a `str`.

PR #1239, opened 2026-03-08
fix(qwen-vl): initialize fps for qwen3_vl and qwen2_5_vl chat path

### LINKED ISSUE #1238: [Bug] "qwen3_vl" & "qwen2_5_vl" chat path crashes with missing fps attribute
## Summary
Running `--model qwen3_vl` can crash on the chat path with:

```
AttributeError: Qwen3_VL object has no attribute fps
```

This happens in `lmms_eval/models/chat/qwen3_vl.py` when `generate_until()` reads `self.fps`, but `fps` is never initialized in the inherited base class (`models/simple/qwen3_vl.py`).

I verified this on current `upstream/main` at commit `ef732077`.

## Minimal Repro
Using multi-GPU (Accelerate) or single-GPU, run a Qwen3-VL eval that goes through the chat implementation, e.g.:

```bash
accelerate launch --num_processes=8 -m lmms_eval \
  --model qwen3_vl \
  --model_args pretrained=Qwen/Qwen3-VL-2B-Instruct,max_num_frames=32 \
  --tasks vsibench \
  --batch_size 1
```

Observed traceback (key frames):

```text
File "lmms_eval/models/chat/qwen3_vl.py", line 64, in generate_until
  if self.fps is not None:
AttributeError: Qwen3_VL object has no attribute fps
```

## Root Cause
`qwen3_vl` resolves to chat implementation by default when both chat/simple are available:

```python
# lmms_eval/models/registry_v2.py
if force_simple and manifest.simple_class_path is not None:
    model_type = "simple"
elif manifest.chat_class_path is not None:
    model_type = "chat"
```

`qwen3_vl` is registered in chat map:

```python
# lmms_eval/models/__init__.py
AVAILABLE_CHAT_TEMPLATE_MODELS = {
    ...
    "qwen3_vl": "Qwen3_VL",
    "qwen2_5_vl": "Qwen2_5_VL",
    ...
}
```

But chat code accesses `self.fps` directly:

```python
# lmms_eval/models/chat/qwen3_vl.py
video_kwargs = {"max_pixels": self.max_pixels, "min_pixels": self.min_pixels}
if self.fps is not None:
    video_kwargs["fps"] = self.fps
    video_kwargs["max_frames"] = self.max_num_frames
else:
    video_kwargs["nframes"] = self.max_num_frames
```

While the inherited simple class does not define `fps` in `__init__`:

```python
# lmms_eval/models/simple/qwen3_vl.py
def __init__(
    self,
    pretrained: str = "Qwen/Qwen3-VL-4B-Instruct",
    device: Optional[str] = "cuda",
    device_map: Optional[str] = "auto",
    batch_size: Optional[Union[int, str]] = 1,
    use_cache=True,
    attn_implementation: Optional[str] = None,
    min_pixels: int = 256 * 28 * 28,
    max_pixels: int = 1605632,
    max_num_frames: int = 32,
    ...
):
    ...
```

I also noticed the same pattern in `lmms_eval/models/chat/qwen2_5_vl.py` (`if self.fps is not None`) while `models/simple/qwen2_5_vl.py` similarly does not initialize `fps`.

Luodian:
## Summary
- initialize `fps` in the shared simple Qwen VL base classes
- guard chat video kwargs construction with `getattr(self, "fps", None)`
- fix the crash reported in #1238 when default model resolution picks the chat implementation

## Validation
- `python3 -m py_compile lmms_eval/models/simple/qwen3_vl.py lmms_eval/models/simple/qwen2_5_vl.py lmms_eval/models/chat/qwen3_vl.py lmms_eval/models/chat/qwen2_5_vl.py`

Closes #1238

### DIFF
--- a/lmms_eval/models/chat/qwen2_5_vl.py
+++ b/lmms_eval/models/chat/qwen2_5_vl.py
@@ -66,9 +66,8 @@ class Qwen2_5_VL(Qwen2_5_VLSimple):
                 "max_pixels": self.max_pixels,
                 "min_pixels": self.min_pixels,
             }
-            fps = getattr(self, "fps", None)
-            if fps is not None:
-                video_kwargs["fps"] = fps
+            if self.fps is not None:
+                video_kwargs["fps"] = self.fps
             else:
                 # Probe videos to get frame count and set nframes = min(max_num_frames, total_frames)
                 # This avoids the error when video has fewer frames than max_num_frames
--- a/lmms_eval/models/chat/qwen3_vl.py
+++ b/lmms_eval/models/chat/qwen3_vl.py
@@ -61,9 +61,8 @@ class Qwen3_VL(Qwen3_VLSimple):
                 "max_pixels": self.max_pixels,
                 "min_pixels": self.min_pixels,
             }
-            fps = getattr(self, "fps", None)
-            if fps is not None:
-                video_kwargs["fps"] = fps
+            if self.fps is not None:
+                video_kwargs["fps"] = self.fps
                 # limit the number of frames in case fps is set
                 video_kwargs["max_frames"] = self.max_num_frames
             else:


Luodian:
@kcz358 could you take a quick look when you have a moment? This draft fixes the missing `fps` initialization behind the `qwen3_vl` / `qwen2_5_vl` chat-path crash reported in #1238.

Main fix is adding `fps` to the shared base init; I also kept a small defensive guard in the chat layer to avoid `AttributeError` if an older path slips through.

kcz358:
Can remove the defensive lines of code. We aren't overriding super init method now so these lines of code are unnecessary but making more complexity and a bit too safe guarding

Luodian:
Updated — I removed the redundant defensive guard and kept the fix scoped to adding `fps` in the shared base init.

PR #1247, opened 2026-03-10
fix: cross-class filelock deadlock in datasets loading

Luodian:
## Summary
- Patch `FileLockMeta.__call__` with a **cross-class singleton cache** so `filelock.FileLock` and `datasets.utils._filelock.FileLock` return the same instance for the same lock path
- Fixes `RuntimeError: Deadlock` during `datasets.load_dataset()` in distributed eval — filelock 3.25.0's global `_registry` detects deadlocks across all classes, but `is_singleton` caches per-class, so two different classes targeting the same `.lock` file trigger a false deadlock
- Adds regression test covering both same-class and cross-class singleton scenarios

## Test plan
- [x] Cross-class singleton verified inside container (filelock 3.25.0 + datasets 4.x)
- [x] 31-task loading passes in single-process container test
- [x] Job 2451 running on 4-node dp32 FluidStack — passed task loading phase (previously deadlocked at this stage)

### DIFF
--- a/lmms_eval/caching/response_cache.py
+++ b/lmms_eval/caching/response_cache.py
@@ -27,9 +27,9 @@ import hashlib
 import inspect
 import json
 import os
-import sqlite3
-import socket
 import shutil
+import socket
+import sqlite3
 import time
 import urllib.parse
 import uuid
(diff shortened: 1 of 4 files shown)


Luodian:
Closing in favor of #1253, which is a strict superset of this PR (same 7 commits + additional SGLang refactor, distributed eval fixes, and cache simplification). All changes from this PR are included there.

PR #1277, opened 2026-03-27
Fix the incompatibility issue caused by `top_p=0` when using vllm to inference (#1265)

### LINKED ISSUE #1265: [Bug] vLLM backend crashes with "top_p must be in (0, 1]" when task config sets top_p=0
### Checklist

## Bug Description

When running evaluation with the `vllm` backend, tasks that set `top_p: 0` in their
`generation_kwargs` (e.g. `pope_random`) cause a crash with:

```
vllm.exceptions.VLLMValidationError: top_p must be in (0, 1], got 0. (parameter=top_p, value=0)
```

The `top_p: 0` setting is commonly used alongside `temperature: 0` and `do_sample: false`
to indicate greedy decoding. This is valid for HuggingFace Transformers, but vLLM's
`SamplingParams` rejects `top_p=0` as out of range.

At least 24 task configs in the repository use `top_p: 0`, including `pope_random`,
`pope_adv`, `pope_pop`, `olympiadbench`, `mathverse`, `mme_cot`, `mmsearch`, etc.
None of these tasks can run with the vLLM backend currently.

## Root Cause

The vLLM wrapper (`lmms_eval/models/simple/vllm.py` and `lmms_eval/models/chat/vllm.py`)
passes `top_p` from task generation_kwargs directly to vLLM's `SamplingParams` without
validation. vLLM requires `top_p` to be in `(0, 1]`, but many task configs set `top_p: 0`
(which is valid for HuggingFace but not for vLLM).

## Affected Files

- `lmms_eval/models/simple/vllm.py` — `generate_until()` method
- `lmms_eval/models/chat/vllm.py` — `make_one_request()` method

### Steps to Reproduce

```shell
pip install -e .
pip install vllm

python -m lmms_eval \
  --model vllm \
  --model_args model=Qwen/Qwen3.5-9B,tensor_parallel_size=1,gpu_memory_utilization=0.85,max_model_len=4096 \
  --tasks pope_random \
  --batch_size 10 \
  --limit 10
```

### Error Message / Traceback

```shell
File "lmms_eval/models/chat/vllm.py", line 107, in generate_until
    sampling_params = SamplingParams(**sampling_params)
File "vllm/sampling_params.py", line 401, in __post_init__
    self._verify_args()
File "vllm/sampling_params.py", line 444, in _verify_args
    raise VLLMValidationError(...)
vllm.exceptions.VLLMValidationError: top_p must be in (0, 1], got 0. (parameter=top_p, value=0)
```

### Environment

- lmms-eval: v0.7.1 (main branch)
- vLLM: v0.18.0 (also affects v0.17.x)
- Python: 3.13
- Model tested: Qwen/Qwen3.5-9B
- GPU: 48GB

## Suggested Fix
Clamp out-of-range `top_p` values to `1.0` before passing to `SamplingParams`.
`top_p=1.0` effectively disables top-p sampling, which is semantically equivalent
to the intent of `top_p=0` (greedy decoding).

## Contribution
I'd be happy to submit a PR to fix this. I've already implemented and verified the fix
locally — the evaluation completes successfully after clamping invalid `top_p` values.

akawincent:
Fixes #1265.

## Summary

Fix a vLLM backend compatibility issue where task configs that set `top_p: 0` crash during request construction.

This change normalizes `top_p=0` to `top_p=1.0` before building vLLM `SamplingParams`, which preserves the intended greedy decoding behavior while satisfying vLLM's validation rules.

## Root Cause

Several task configs in `lmms-eval` use:

- `temperature: 0`
- `top_p: 0`
- `do_sample: false`

This pattern is commonly used to express greedy decoding and works in Hugging Face-style generation paths.

However, the vLLM wrappers were forwarding `top_p` directly into `SamplingParams`, and vLLM requires `top_p` to be in `(0, 1]`. As a result, tasks using `top_p: 0` failed before generation started.

## What Changed

- Added a small normalization step in the shared vLLM wrapper to convert `top_p=0` to `1.0`
- Reused that logic in:
  - `vllm`
  - `vllm_chat`
  - `vllm_generate`
- Added a lightweight regression test covering the normalization path

## Impact

Tasks that rely on greedy-style generation settings with `top_p: 0` can now run on vLLM without crashing.

This keeps task YAML unchanged and limits the fix to the vLLM compatibility layer.

## Validation

Ran:

- `python -m unittest test.models.test_vllm_sampling_params -v`
- `python -m unittest test.models.test_model_registry_v2 -v`


akawincent:
Hi @Luodian 

Could you review this PR please?

The code changes are minimal: it mainly converts top_p=0 to top_p=1 when calling the vLLM backend, so that vLLM's greedy decoding can start.

Additionally, I added a regression test.

test/models/test_vllm_sampling_params.py
kcz358, line 1: I think this test file does not need to be commited

akawincent: That makes sense. This regression test is only valid for my local env.

akawincent: Just to confirm, should I remove this test file from the PR, or would you update it directly if you have access?

kcz358: Hi, can just remove this test file from the PR. Thanks

akawincent: Hi, I just deleted test file. 😄

kcz358 (approved):

PR #1278, opened 2026-03-27
fix: preserve HME100k prediction case in OCRBench scoring

### LINKED ISSUE #1220: OCRBench eval bug
https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/23bfbf9f5d309d694090d6da7fd6355ff4c28cab/lmms_eval/tasks/ocrbench/utils.py#L34C5-L34C38

这里将pred转小写，但是HME100k分支里的answer没有转小写，导致回答正确但score=0

akawincent:
## Summary
- fix OCRBench scoring for the HME100k subset by preserving prediction case
- keep the existing lowercase normalization for the other OCRBench subsets unchanged

## Why
Issue #1220 points out that `ocrbench_process_results` lowercases `pred` before branching on `dataset_name`, while the `HME100k` branch intentionally compares answers without lowercasing them. That makes correct HME100k predictions score as 0 when the only difference is letter case.

## Testing
- not run (per request)

Closes #1220

### DIFF
--- a/lmms_eval/tasks/ocrbench/utils.py
+++ b/lmms_eval/tasks/ocrbench/utils.py
@@ -31,7 +31,7 @@ def ocrbench_doc_to_text(doc, lmms_eval_specific_kwargs):
 
 
 def ocrbench_process_results(doc, results):
-    pred = results[0].lower().strip()
+    pred = results[0].strip()
     gt_ans = doc["answer"]
     dataset_name = doc["dataset"]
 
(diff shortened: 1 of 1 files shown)


kcz358 (commented):
Hi, so instead of just remove the lower(), maybe should actually lower the gt_ans as well?

akawincent:
> Hi, so instead of just remove the lower(), maybe should actually lower the gt_ans as well?

@kcz358  I don't think we should lowercase `gt_ans` for HME100k.

HME100k should be case-sensitive, since it is handwritten mathematical expression recognition.

So, we should not lowercase `pred`, and we should not lowercase `gt_ans` either. The original problem was that `pred` was lowercased too early, which broke the intended HME100k matching behavior.

Since these answers are math-expression / LaTeX-like strings, lowercasing `gt_ans` could also create false positives for charactors like V, F, I, A, etc.

kcz358:
Got it, looks make sense if HME requires the answer to be case sensitive. Will this change cause false negative on other branches? If not then I will merge this PR. Thanks

akawincent:
@kcz358 

No, this should not affect the other OCRBench branches. 
I have confirmed that other branches do  `pred.lower()` and `gt_ans.lower()` in `else....` cuz they are **case-insensitive**.

kcz358 (approved):

PR #1279, opened 2026-03-28
fix: handle internvl_hf video-only inputs and enable frame sampling

### LINKED ISSUE #1241: [Bug] internvl_hf: IndexError when visuals is empty list (video-only inputs)
### Checklist

## Bug Description

When running `internvl_hf` with video-only tasks (no images), 
`generate_until` passes an empty list `[]` to `self.processor(images=visuals, ...)`, 
which causes an `IndexError` inside the processor.

## Error Traceback
```
File "lmms_eval/models/chat/internvl_hf.py", line 270, in generate_until
    inputs = self.processor(images=visuals, ...)
...
File "transformers/image_transforms.py", line 917, in group_images_by_shape
    device = images[0][0].device if is_nested else images[0].device
IndexError: list index out of range
```

## Root Cause

`videos` is already handled with a `None` guard, but `visuals` and `image_sizes` are not:
```python
# Only videos is guarded — visuals is not
if len(videos) == 0:
    videos = None

# This also crashes when visuals is []
gen_kwargs["image_sizes"] = [visuals[idx].size for idx in range(len(visuals))]
```

## Fix
```python
# Before
if len(videos) == 0:
    videos = None

inputs = self.processor(
    images=visuals,
    ...
)

gen_kwargs["image_sizes"] = [visuals[idx].size for idx in range(len(visuals))]
```
```python
# After
if len(videos) == 0:
    videos = None
if len(visuals) == 0:
    visuals = None

inputs = self.processor(
    images=visuals,
    ...
)

gen_kwargs["image_sizes"] = [visuals[idx].size for idx in range(len(visuals))] if visuals is not None else []
```

## Environment

- lmms-eval version: main branch
- Model: `internvl_hf` with `OpenGVLab/InternVL3-8B-hf`
- Task type: video-only (no image inputs)

### Steps to Reproduce

```shell
.
```

### Error Message / Traceback

```shell
.
```

### Environment

.

### Additional Context

_No response_

### LINKED ISSUE #1242: [Bug] Internvl_hf: num_frames not applied to videos
### Checklist

### Root Cause
`InternVLVideoProcessor` has `do_sample_frames = False` as default,
so `num_frames` passed via `videos_kwargs` is completely ignored.
All frames in the video are used regardless of the `num_frames` argument,
causing context length overflow:
```
Token indices sequence length is longer than the specified maximum sequence length
for this model (11912 > 8192). Running this sequence through the model will result
in indexing errors
```

### Fix
```python
# Before
videos_kwargs = {}
if self.num_frames is not None:
    videos_kwargs["num_frames"] = self.num_frames

# After
videos_kwargs = {}
videos_kwargs["do_sample_frames"] = True  # must explicitly enable frame sampling
if self.num_frames is not None:
    videos_kwargs["num_frames"] = self.num_frames
```

---
### Additional Context

The root cause can be confirmed at:
https://github.com/huggingface/transformers/blob/fefc3fa205aae3fb44127142e0771b0c8762ee46/src/transformers/models/internvl/video_processing_internvl.py#L43

### Steps to Reproduce

```shell
"launch",
            "--main_process_port","43812",
            "--num_processes=8", 
            "-m", "lmms_eval",
            "--model", "internvl_hf",
            "--model_args", "pretrained=OpenGVLab/InternVL3-8B-hf,num_frames=8,max_patches=1",
            "--tasks", "$TASKS",
            "--batch_size", "1",
            "--log_samples",
            "--output_path", "./logs/internvl_3-8b",
      ],
```

### Error Message / Traceback

```shell
Token indices sequence length is longer than the specified maximum sequence length
for this model (11912 > 8192). Running this sequence through the model will result
in indexing errors
.
```

## Environment
- File: `lmms_eval/models/chat/internvl_hf.py`
- Model: `internvl_hf` with `OpenGVLab/InternVL3-8B-hf`
- Task type: video-only (no image inputs)

### Additional Context

_No response_

akawincent:
## Summary

This PR fixes two `internvl_hf` video handling bugs in `lmms_eval/models/chat/internvl_hf.py`:

- explicitly enables frame sampling when `num_frames` or `fps` is configured
- normalizes empty image inputs to `None` for video-only examples
- avoids building `image_sizes` from an empty image list

## Why

This PR is motivated by two related `internvl_hf` bugs reported upstream:

- #1241 reports that video-only inputs can crash because `generate_until()` passes `images=[]` into the processor, and later also assumes `visuals` is non-empty when building `image_sizes`
- #1242 reports that `num_frames` is not actually applied, because InternVL video processing keeps `do_sample_frames=False` by default unless it is explicitly enabled

Together, these two issues make `internvl_hf` unreliable on video tasks:
video-only samples can raise an `IndexError`, and sampled-video runs can still process all frames and overflow the model context length.

Closes #1241  
Closes #1242

## Validation

- `pre-commit run --files lmms_eval/models/chat/internvl_hf.py`
- `python -m py_compile lmms_eval/models/chat/internvl_hf.py`

### DIFF
--- a/lmms_eval/models/chat/internvl_hf.py
+++ b/lmms_eval/models/chat/internvl_hf.py
@@ -263,6 +263,8 @@ class InternVLHf(lmms):
             if self.accelerator.is_main_process and doc_id[0] % 100 == 0:
                 eval_logger.debug(f"Prompt for doc ID {doc_id[0]}:\n\n{text}\n")
 
+            if len(visuals) == 0:
+                visuals = None
             if len(videos) == 0:
                 videos = None
             inputs = self.processor(
@@ -278,7 +280,7 @@ class InternVLHf(lmms):
             # this is safe to assume because the `grouper` object ensures it.
             gen_kwargs = all_gen_kwargs[0]
 
-            gen_kwargs["image_sizes"] = [visuals[idx].size for idx in range(len(visuals))]
+            gen_kwargs["image_sizes"] = [visual.size for visual in visuals] if visuals is not None else []
             if "max_new_tokens" not in gen_kwargs:
                 gen_kwargs["max_new_tokens"] = 1024
             if "temperature" not in gen_kwargs:
--- a/lmms_eval/models/chat/internvl_hf.py
+++ b/lmms_eval/models/chat/internvl_hf.py
@@ -249,6 +249,9 @@ class InternVLHf(lmms):
                 images_kwargs["min_patches"] = self.min_patches
             if self.max_patches is not None:
                 images_kwargs["max_patches"] = self.max_patches
+            if self.num_frames is not None or self.fps is not None:
+                # InternVL only applies num_frames/fps when frame sampling is explicitly enabled.
+                videos_kwargs["do_sample_frames"] = True
             if self.num_frames is not None:
                 videos_kwargs["num_frames"] = self.num_frames
             if self.fps is not None:


akawincent:
Hi @Luodian!  

Need for review and tests.
InternVL is a widely used model, so I think this issue should be resolved as soon as possible.

mwxely (commented):
Clean and well-scoped fix. Two minor points to consider:

1. **Implicit behavioral change**: Since `num_frames` defaults to `32` in `__init__`, setting `do_sample_frames=True` means frame sampling will now **actually take effect** by default. Previously `num_frames=32` was silently ignored (since `do_sample_frames` was never set). This is arguably the correct behavior, but worth noting in the PR description as it changes the default video processing pipeline for all `internvl_hf` users on video tasks.

2. **Validation question**: Did you run an end-to-end video eval (e.g., a small subset of VideoMME or similar) to confirm the fix works beyond `py_compile` and `pre-commit`? A quick `--limit 4` smoke test would be reassuring.

Otherwise LGTM — the `visuals = None` guard and `image_sizes` fix are correct and consistent with existing patterns in the file. cc @kcz358

kcz358 (approved):

akawincent:
> 2. **Validation question**: Did you run an end-to-end video eval (e.g., a small subset of VideoMME or similar) to confirm the fix works beyond `py_compile` and `pre-commit`? A quick `--limit 4` smoke test would be reassuring.
> 

@mwxely Hello,

Yes, I did.  Actually, I ran it with `--limit` to test whether the code works, but I didn't run a full benchmark. To be rigorous, a better testing would be to change the `num_frames` and observe changes in the final metrics.

PR #1298, opened 2026-04-14
feat: add Bedrock and local vLLM providers for llm_judge

ShownX:
- BedrockProvider: AWS Bedrock via boto3 Converse API, supports bearer token auth
- LocalProvider: any OpenAI-compatible local server (vLLM, SGLang)
- Register both in ProviderFactory
- Load .env via python-dotenv for configuration


kcz358 (approved):
Sorry for the late review, LGTM. Thanks for the contributions.

PR #1301, opened 2026-04-17
Add WISE Benchmark Task

Purshow:
## Summary

- 
- 
- 

## In scope

- 

## Out of scope

- 

## Validation

- 

## Risk / Compatibility

- 

## Type of Change
- [ ] New feature
- [ ✅] New benchmark/task
- [ ] New model integration
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/lmms_eval/tasks/WISE/WISE.yaml
+++ b/lmms_eval/tasks/WISE/WISE.yaml
@@ -2,7 +2,7 @@
 
 dataset_path: Yuwei-Niu/WISE
 task: WISE
-test_split: train
+test_split: test
 output_type: generate_until
 
 doc_to_visual: !function utils.wise_doc_to_visual
--- a/lmms_eval/tasks/WISE/WISE.yaml
+++ b/lmms_eval/tasks/WISE/WISE.yaml
@@ -1,15 +1,17 @@
 # WISE text-to-image benchmark.
 #
 # Dataset notes:
-# - The source JSON is expected to be uploaded to Hugging Face as:
-#     Yuwei-Niu/WISE/final_data.json
-# - data_files maps that JSON into the lmms-eval test split without using any
-#   local mount path.
+# - The official WISE release stores prompts in three JSON files.
+# - data_files maps those files into the lmms-eval test split without using any
+#   local mount path or invalid HuggingFace repo-as-file value.
 
 dataset_path: json
 dataset_kwargs:
   data_files:
-    test: Yuwei-Niu/WISE
+    test:
+      - https://raw.githubusercontent.com/PKU-YuanGroup/WISE/main/data/cultural_common_sense.json
+      - https://raw.githubusercontent.com/PKU-YuanGroup/WISE/main/data/spatio-temporal_reasoning.json
+      - https://raw.githubusercontent.com/PKU-YuanGroup/WISE/main/data/natural_science.json
 task: WISE
 test_split: test
 output_type: generate_until
--- a/lmms_eval/tasks/WISE/WISE.yaml
+++ b/lmms_eval/tasks/WISE/WISE.yaml
@@ -1,19 +1,8 @@
 # WISE text-to-image benchmark.
-#
-# Dataset notes:
-# - The official WISE release stores prompts in three JSON files.
-# - data_files maps those files into the lmms-eval test split without using any
-#   local mount path or invalid HuggingFace repo-as-file value.
 
-dataset_path: json
-dataset_kwargs:
-  data_files:
-    test:
-      - https://raw.githubusercontent.com/PKU-YuanGroup/WISE/main/data/cultural_common_sense.json
-      - https://raw.githubusercontent.com/PKU-YuanGroup/WISE/main/data/spatio-temporal_reasoning.json
-      - https://raw.githubusercontent.com/PKU-YuanGroup/WISE/main/data/natural_science.json
+dataset_path: Yuwei-Niu/WISE
 task: WISE
-test_split: test
+test_split: train
 output_type: generate_until
 
 doc_to_visual: !function utils.wise_doc_to_visual
--- a/lmms_eval/tasks/WISE/WISE.yaml
+++ b/lmms_eval/tasks/WISE/WISE.yaml
@@ -1,0 +1,60 @@
+# WISE text-to-image benchmark.
+#
+# Dataset notes:
+# - The source JSON is expected to be uploaded to Hugging Face as:
+#     Yuwei-Niu/WISE/final_data.json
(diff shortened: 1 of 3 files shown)


mwxely (commented):
## Review: Add WISE Benchmark Task

I pushed a lint fix (black + isort formatting) in commit `159706b` to get CI green.

### Issues Found

**1. Dataset loading config looks wrong**
```yaml
dataset_path: json
dataset_kwargs:
  data_files:
    test: Yuwei-Niu/WISE
```
The `json` dataset loader expects a file path or URL to a JSON file, not a HuggingFace dataset repo identifier. This will likely fail at runtime. Should be either:
- `dataset_path: Yuwei-Niu/WISE` (if it's a proper HF dataset), or
- Provide the actual URL to the JSON file (e.g., `https://huggingface.co/datasets/Yuwei-Niu/WISE/resolve/main/final_data.json`)

**2. `output_type: generate_until` vs text-to-image**
This task evaluates *generated images*, but `generate_until` produces text. The code in `_find_generated_image` looks for images saved to `WISE_RAW_OUTPUT_DIR` by the model, but the connection between lmms-eval's generation pipeline and image output is unclear. A brief explanation in the README or PR description would help.

**3. `wise_process_results` relies on `doc_id` from `**kwargs`**
```python
doc_id = kwargs.get("doc_id", 0)
```
Does lmms-eval's `process_results` callback actually pass `doc_id` in kwargs? If not, every sample will use `doc_id=0` and fail to find its image. This needs verification.

**4. Hardcoded local paths in README**
```
cd /pfs/weiyang/Show/lmms-eval
```
and
```
WISE_RAW_OUTPUT_DIR="/pfs/weiyang/Show/lmms-eval/outputs/WISE_raw/bagel_umm"
```
These are internal paths. Please replace with generic placeholders.

**5. `_find_generated_image` is model-specific**
The function hardcodes naming patterns for `bagel.py`, `bagel_unig2u.py`, `mmada.py`. This is fragile — any new model integration would require modifying this function. Consider a more generic approach (e.g., glob for `*{doc_id}*` or let the model specify the naming convention).

### Summary
The core evaluation logic (GPT-4o judge → 3 scores → WiScore) is sound and well-structured. However, there are several integration concerns (dataset loading, doc_id passing, image discovery) that need verification before merge.

lmms_eval/tasks/WISE/WISE.yaml
```diff
@@ -0,0 +1,60 @@
...
+#     Yuwei-Niu/WISE/final_data.json
+# - data_files maps that JSON into the lmms-eval test split without using any
+#   local mount path.
+
+dataset_path: json
+dataset_kwargs:
+  data_files:
+    test: Yuwei-Niu/WISE
+task: WISE
+test_split: test
```
kcz358: If can use `load_dataset`, then can use a better format in my opinion. Maybe we can try just use `Yuwei-Niu/WISE` ?

lmms_eval/tasks/WISE/utils.py
```diff
@@ -0,0 +1,610 @@
...
+  2. Model generates images to output_dir (e.g., WISE_raw/bagel_umm/).
+  3. Find generated images from output_dir based on doc_id.
+  4. Use an OpenAI-compatible multimodal judge to produce three scores:
+     consistency, realism, aesthetic_quality (each 0-2).
+  5. Calculate WiScore = (0.7*consistency + 0.2*realism + 0.1*aesthetic) / 2.
+  6. Aggregate WiScores by WISE category weights.
+
+Environment variables:
+  - WISE_RAW_OUTPUT_DIR: model output directory (e.g., /pfs/.../WISE_raw/bagel_umm).
+    Required for finding generated images.
```
kcz358: This does not looks good in my opinion. Should refer to https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/main/lmms_eval/models/simple/bagel.py and https://github.com/EvolvingLMMs-Lab/lmms-eval/blob/main/lmms_eval/models/chat/bagel_lmms_engine.py to direct get image path from answer. Maybe can also refer to other task such as imgedit. Thanks!

lmms_eval/tasks/WISE/utils.py
```diff
@@ -0,0 +1,610 @@
...
+        f"WISE_{doc_id}_0.png",  # bagel.py
+        f"WISE_{doc_id}.png",  # bagel_unig2u, mmada
+    ]
+
+    for name in possible_names:
+        path = os.path.join(search_dir, name)
+        if os.path.exists(path):
+            return path
+
+    return None
```
kcz358: After addressing the env var set for wise, I think this part would not be needed as we get image path direct from output.

Purshow:
Thanks for the review. I addressed the WISE integration issues:

- Replaced the huggingface json data_files entry with actual official WISE JSON URLs.
- Removed WISE_RAW_OUTPUT_DIR and model-specific image discovery.
- WISE now follows the imgedit-style contract: model wrappers return JSON with an images list, and process_results reads images[0].
- Removed doc_id dependency from process_results; prompt_id/category now come from the dataset doc.
- Replaced internal README paths with generic placeholders.
- Reduced sample logging duplication by keeping detailed per-sample info only under WISE_overall_wiscore and using compact scalar category metrics.

kcz358:
Hi, I think the github content is not what we wish to use here. Just like imgedit that we can use

```yaml
dataset_path: kcz358/imgedit
dataset_kwargs:
  token: True
task: "imgedit"
test_split: test
```
so that is equivalent to `load_dataset(kcz358/imgedit, split=test)`. Since you use https://huggingface.co/datasets/Yuwei-Niu/WISE, I think you can put dataset path as Yuwei-Niu/WISE and test split as `train` to use the huggingface auto generate one. If you want to set it properly, check https://huggingface.co/docs/hub/datasets-manual-configuration for info on setting the files and split manually.

Purshow:
> Hi, I think the github content is not what we wish to use here. Just like imgedit that we can use
> 
> ```yaml
> dataset_path: kcz358/imgedit
> dataset_kwargs:
>   token: True
> task: "imgedit"
> test_split: test
> ```
> 
> so that is equivalent to `load_dataset(kcz358/imgedit, split=test)`. Since you use https://huggingface.co/datasets/Yuwei-Niu/WISE, I think you can put dataset path as Yuwei-Niu/WISE and test split as `train` to use the huggingface auto generate one. If you want to set it properly, check https://huggingface.co/docs/hub/datasets-manual-configuration for info on setting the files and split manually.

I have addressed it. Thx!

kcz358 (approved):
LGTM, thanks for the contributions!

PR #1307, opened 2026-04-24
feat: add ReVSI evaluation

eamonn-zh:
## Summary
- Add evaluation code for [ReVSI benchmark](https://3dlg-hcvc.github.io/revsi/)

## In scope
- Add a new task revsi

## Type of Change
- [ ] New feature
- [x] New benchmark/task
- [ ] New model integration
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/lmms_eval/tasks/revsi/_default_template_yaml
+++ b/lmms_eval/tasks/revsi/_default_template_yaml
@@ -1,0 +1,27 @@
+dataset_path: 3dlg-hcvc/ReVSI
+test_split: test
+dataset_kwargs:
+  token: True
+  cache_dir: revsi
+  video: True
+output_type: generate_until
+process_docs: !function utils.process_docs
+doc_to_visual: !function utils.revsi_doc_to_visual
+doc_to_text: !function utils.revsi_doc_to_text
+doc_to_target: "ground_truth"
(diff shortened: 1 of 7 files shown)


kcz358 (commented):
Hi, thank you for the contributions. For the different frames, I think this should be model parameters adjustment instead of creating extra tasks? So maybe we should only need the all frame yaml if that is the case. And the the aggregation, it would be better if you can add the metrics in the metrics list so it will be logged in the result.json instead just use logger to log the results. Thank you.

eamonn-zh:
Hi @kcz358, thanks for the suggestions! We have added the metrics to the metrics list and removed the self-logging of results.

Regarding the different frame settings, [ReVSI](https://3dlg-hcvc.github.io/revsi/) defines four frame-budget subsets. **Each subset provides different ground-truth answers to ensure that questions remain answerable under the corresponding frame sampling budget** and are not invalidated by missing visual evidence. This frame-adaptive evaluation protocol is an important feature and core contribution of ReVSI, so all four sub-tasks (all-frame, 64-frame, 32-frame and 16-frame) are necessary. Thank you!

kcz358 (approved):
Thanks for the clarification. LGTM

PR #1320, opened 2026-05-02
fix: preserve text task context in OpenAI chat adapter

### LINKED ISSUE #1319: OpenAI chat adapter drops text-task ctx, causing lm-eval-harness task mismatch
### Checklist

### Bug Description

Hi, we found a behavior mismatch when running text-only tasks such as `mmlu_pro` through `lmms-eval` with the OpenAI-compatible adapter.

`lmms-eval` builds the full task context `ctx` correctly, including the task description, few-shot context, and current question. However, the OpenAI chat adapter appears to ignore that constructed `ctx` and rebuilds messages from `doc_to_messages`.

Relevant path:

```python
# lmms_eval/api/task.py
arguments = (
    ctx,
    self.doc_to_messages,
    copy.deepcopy(self.config.generation_kwargs),
    doc_id,
    self.config.task,
    split,
)
```

But in the OpenAI chat adapter:

```python
# lmms_eval/models/chat/openai.py
_, doc_to_messages, gen_kwargs, doc_id, task, split = req.args
chat_messages_raw = doc_to_messages(self.task_dict[task][split][doc_id])
```

The first arg, `ctx`, is discarded.

For `mmlu_pro`, this means the actual request sent to the backend omits the task description:

```text
The following are multiple choice questions (with answers) about {subject}.
Think step by step and then finish your answer with "the answer is (X)"
where X is the correct letter choice.
```

This causes benchmark behavior to diverge from `lm-eval-harness`, because the model is no longer instructed to produce the answer format expected by `custom-extract`.

## Expected behavior

For text-only tasks inherited from `lm-eval-harness`, the OpenAI-compatible backend request should preserve the same constructed prompt semantics as `lm-eval-harness`. Either:

- use `ctx` directly for text-only tasks,
- ensure `doc_to_messages` includes the same description/few-shot context, or
- route text-only tasks through the simple adapter by default.

## Actual behavior

The request payload contains only the question message, without the task description/system prompt. This changes model output format and measured accuracy.

We also observed related compatibility issues:

- `--gen_kwargs` is parsed, but the OpenAI adapter forwards only a subset, e.g. `max_new_tokens` and `temperature`, dropping params such as `top_p`, `top_k`, `min_p`, penalties, and `until`/`stop`.
- The OpenAI adapter applies a hard `4096` cap to `max_new_tokens`, which is problematic for local OpenAI-compatible backends.
- `RegexFilter` can crash when a regex alternation returns an all-empty capture tuple:

```python
match = [m for m in match if m][0]
# IndexError if all groups are empty
```

These differences make `lmms-eval` non-equivalent to `lm-eval-harness` for text benchmarks despite sharing task names.

A minimal fix for the main issue may be to have the OpenAI chat adapter use the constructed `ctx` for text-only tasks, or force the simple adapter for text-only tasks where multimodal structured messages are not required.

### Steps to Reproduce

```shell
NA
```

### Error Message / Traceback

```shell

```

### Environment

NA

### Additional Context

_No response_

Genmin:
## Summary
- preserve the constructed `ctx` for text-only OpenAI chat requests instead of sending only the raw `doc_to_messages` question text
- keep explicit multi-message chats and multimodal messages on the existing `doc_to_messages` path
- add regression coverage for text-only replacement plus multimodal/chat-native preservation

Fixes #1319.

## Validation
- `uv run --with pytest --with loguru --with python-dotenv --with openai --with pydantic --with tqdm --with numpy --with pillow --with accelerate python -m pytest test/models/test_openai_chat.py -q`
- `uv run --with loguru --with python-dotenv --with openai --with pydantic --with tqdm --with numpy --with pillow --with accelerate python -m py_compile lmms_eval/models/chat/openai.py test/models/test_openai_chat.py`
- `git diff --check`

### DIFF
--- a/lmms_eval/models/chat/openai.py
+++ b/lmms_eval/models/chat/openai.py
@@ -30,6 +30,29 @@ cpu, _ = optional_import("decord", "cpu")
 load_dotenv(verbose=True)
 
 
+def _content_parts_are_text_only(content) -> bool:
+    if isinstance(content, str):
+        return True
+    if not isinstance(content, list):
+        return False
+    return all(isinstance(part, dict) and part.get("type") == "text" for part in content)
+
+
+def _ctx_to_text_chat_messages(ctx: str, chat_messages_raw):
+    if not isinstance(ctx, str) or not ctx:
+        return chat_messages_raw
+    if not isinstance(chat_messages_raw, list) or len(chat_messages_raw) != 1:
+        return chat_messages_raw
+
+    message = chat_messages_raw[0]
+    if not isinstance(message, dict) or message.get("role") != "user":
+        return chat_messages_raw
+    if not _content_parts_are_text_only(message.get("content")):
+        return chat_messages_raw
+
+    return [{"role": "user", "content": [{"type": "text", "text": ctx}]}]
+
+
 @register_model("openai")
 class OpenAICompatible(OpenAICompatibleSimple):
     is_simple = False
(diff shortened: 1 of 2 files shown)


kcz358 (commented):
Hi, I think this would hurt most of the tasks that use doc to messages to construct the messages. If there are tasks with wrong auto wrapped doc to messages function, the tasks should be revise instead of changing the doc to messages logic in the oai model.

Genmin:
Thanks for taking a look. I agree with the concern: this is too broad for the model adapter layer, and task-specific message construction should be fixed at the task level instead. Closing this to avoid pushing behavior that could regress normal doc_to_messages tasks.

PR #1330, opened 2026-05-12
fix(jumpscore): align message format and video lookup

mathCrazyy:
## Summary

- Remove legacy count QA context from `doc_to_messages` so evaluation input matches the intended timestamp-only prompt.

## In scope

- Update `lmms_eval/tasks/jumpscore/utils.py`.
- Change JumpScore chat message construction from multi-turn count QA context to a single-turn timestamp query.

## Out of scope

- No model implementation changes.
- No metric or scoring logic changes.
- No dataset content changes.
- No changes to prompts outside JumpScore.
- No changes to other tasks.

## Validation

- Confirmed the committed diff only touches `lmms_eval/tasks/jumpscore/utils.py`.
- Verified the message construction now produces one user turn containing the video and timestamp question.
- Verified the video lookup includes existing cache paths plus HF snapshot fallback paths.
- Pushed commit `cf7f49f` to `origin/main`.

## Risk / Compatibility

- Low risk for model code because this only changes JumpScore task input construction.
- Expected behavior change for JumpScore evaluation prompts: legacy count QA history is no longer included.
- Compatible with existing cache layouts; adds support for HF snapshot cache layout as a fallback.
- Results may differ from prior JumpScore runs because the evaluation input format is now aligned to the single-turn protocol.

## Type of Change
- Evaluation/task configuration alignment


lmms_eval/tasks/jumpscore/utils.py
```diff
@@ -24,6 +24,16 @@ def jumpscore_doc_to_visual(doc: Dict[str, Any], lmms_eval_specific_kwargs: Opti
...
+        snapshot_root = os.path.join(hf_home, "hub", "datasets--lmms-lab-encoder--JumpScore", "snapshots")
+        if os.path.isdir(snapshot_root):
+            for snapshot in os.listdir(snapshot_root):
+                snapshot_dir = os.path.join(snapshot_root, snapshot)
+                candidates.extend(
+                    [
+                        os.path.join(snapshot_dir, video_ref),
+                        os.path.join(snapshot_dir, "videos", video_ref),
+                    ]
+                )
```
kcz358: The format of the jumpscore is a bit not follow the convention of other video tasks. Downloading many small files from huggingface is not efficient. Consider to zip the videos into zip with each shard lower than 5GB and unzip locally or prepared locally into the cache dir. This fallback should be remove.

kcz358:
JumpScore does not zip the data yet. Will it be zipped later? If this is the case, I will merge this PR. Thanks!

mathCrazyy:
> JumpScore does not zip the data yet. Will it be zipped later? If this is the case, I will merge this PR. Thanks!

Thank you for the review!
The data has been updated to zip format, and I’ve already adapted the code to support it.
Feel free to merge this PR when you’re ready. Thanks!

PR #1361, opened 2026-06-05
Add Qwen-native JSON coordinate variants for pointing tasks

njb-nvidia:
## Summary

Qwen-VL grounding models emit points as a JSON list of `{"point_2d": [x, y], ...}` entries on a 0-1000 normalized grid. The existing pointing tasks prompt for and parse `(x, y)` tuples, so the default parser returns nothing on this native output.

This PR adds opt-in `*_json` task variants that prompt for, and parse, Qwen's native format:

- `where2place_json`
- `refspatial_location_json`, `refspatial_placement_json`, `refspatial_unseen_json`
- `pointbench_json`

Each variant overrides only `doc_to_text` and `process_results` and shares a small `parse_point2d` helper (`lmms_eval/tasks/_task_utils/point_format.py`). The existing tuple tasks, their metrics, and aggregation are **unchanged** (verified: no diff to the existing `doc_to_text`/prompt constants).

Stacks on #1360 — `pointbench_json` relies on that PR's image-alignment fix and binary metric so its parsed points are scored against the correct images.

## Validation

Run against a Qwen3-VL-8B grounding fine-tune (vllm) and compared to a reference fork that uses the same JSON prompt + parser:

| Benchmark | this PR (`*_json`) | reference | Δ |
| --- | --- | --- | --- |
| where2place | 0.6815 | 0.6829 | −0.001 |
| refspatial (size-weighted) | 0.560 | 0.592 | −0.032 |
| pointbench (binary) | 0.677 | 0.632 | +0.045 |

where2place matches almost exactly. The refspatial gap is borderline single-point flips on the harder splits. pointbench runs slightly higher because `parse_point2d` recovers points from mildly malformed JSON that a strict fenced-JSON parser drops, and because the reference injects an extra steerable-category hint not ported here.

### DIFF
--- a/lmms_eval/tasks/_task_utils/point_format.py
+++ b/lmms_eval/tasks/_task_utils/point_format.py
@@ -1,0 +1,31 @@
+"""Shared helper for Qwen-native point parsing.
+
+Qwen-VL grounding models emit points as a JSON list of
+``{"point_2d": [x, y], ...}`` entries on a 0-1000 normalized grid. The pointing
+tasks (where2place / refspatial / pointbench) ship ``*_json`` variants that
+prompt for this format and parse it with :func:`parse_point2d` below.
+"""
+
+import re
+
+import numpy as np
+
+
+def parse_point2d(text: str, width: int, height: int) -> np.ndarray:
+    """Parse ``[x, y]`` / ``(x, y)`` coordinate pairs into pixel points.
+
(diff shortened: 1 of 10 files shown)


kcz358:
Hi, I think this PR has some conflictions with your previous PR on pointbench. Maybe you need to rebase on the main first to solve it. Thanks!

PR #1366, opened 2026-06-16
feat(openai): add pass_video_url and enable_thinking_kwarg for vLLM-served video tasks

min1321:
## Summary

- Add two opt-in init params to `OpenAICompatible`: `pass_video_url` and `enable_thinking_kwarg`, so a vLLM-served Qwen3-VL / Qwen3.5-VL backend can do **server-side** video decoding instead of forcing client-side frame extraction.
- `pass_video_url=True` sends each video as `{"type": "video_url", "video_url": {"url": "file://..."}}` so vLLM can apply `media_io_kwargs.num_frames` and attach absolute-time signals; `enable_thinking_kwarg` forwards `chat_template_kwargs.enable_thinking` via `extra_body`.
- Defaults are unchanged — existing tasks/configs are unaffected.

## In scope

- `lmms_eval/models/chat/openai.py`:
  - `OpenAICompatible.__init__`: accept `pass_video_url: bool = False` and `enable_thinking_kwarg: object = None`.
  - `build_payload_for_index`: when `pass_video_url=True`, build the OpenAI `messages` list manually (skip `to_openai_messages`) and emit each video as a `video_url` part. When either flag is set, populate `payload["extra_body"]` with `media_io_kwargs` and/or `chat_template_kwargs`.

## Out of scope

- Other adapters (`vllm`, `sglang`, `huggingface`, `litellm`, …). Same idea would apply but each has its own video path; happy to follow up if maintainers want.
- Changes to `qwen_vl_utils` or `protocol.py`.
- Audio / image handling — unchanged.

## Validation

- `python -m lmms_eval --model openai --tasks extremewhenbench --model_args "...,pass_video_url=True,max_frames_num=768,enable_thinking_kwarg=False,..."` against vLLM-served Qwen3.5-9B | sample size: `N=2,273` | key metrics: mIoU | result: **0.048 with new flags vs. 0.003 default (existing path) — pass.** Matches a hand-rolled `openai`-client reference (0.047) within run-to-run noise.
- Default-path regression: existing tasks via `--model openai` without the new flags produce identical output | result: **pass**.

## Risk / Compatibility

- Defaults preserve existing behavior; new flags are opt-in. No new dependencies.
- `extra_body` is OpenAI-API spec-compliant; non-vLLM backends that don't understand `media_io_kwargs` simply ignore it (server-specific).

## Type of Change
- [x] New feature
- [ ] New benchmark/task
- [ ] New model integration
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/lmms_eval/models/chat/openai.py
+++ b/lmms_eval/models/chat/openai.py
@@ -35,6 +35,11 @@ load_dotenv(verbose=True)
 class OpenAICompatible(OpenAICompatibleSimple):
     is_simple = False
 
+    def __init__(self, *args, pass_video_url: bool = False, enable_thinking_kwarg: object = None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.pass_video_url = bool(pass_video_url)
+        self.enable_thinking_kwarg = enable_thinking_kwarg
+
     def generate_until(self, requests) -> List[GenerationResult]:
         if not requests:
             return []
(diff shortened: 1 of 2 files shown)


kcz358:
Thanks! Two thoughts:

1. `enable_thinking_kwarg` is fine to keep on the wrapper side.

2. `pass_video_url`: this branch should live in `protocol.to_openai_messages` rather than as a manual messages-builder in `build_payload_for_index`. Add a `pass_video_url: bool = False` arg to `to_openai_messages` in `lmms_eval/protocol.py`; when True the video branch emits `{"type": "video_url", "video_url": {"url": ...}}` (with the `file://` normalization) instead of decoding to base64 frames. Then `openai.py` just forwards the flag — `chat_messages.to_openai_messages(video_kwargs=video_kwargs, pass_video_url=self.pass_video_url)` — and keeps the `extra_body` block. Same hook can later be added to `to_qwen3_vl_openai_messages` for the async wrapper. A nice side effect: the current manual loop forwards raw image URLs as `image_url.url`, dropping the base64/data-URL encoding for local image paths — keeping the transform centralized avoids that regression.

Quick question: which tasks actually need this? Every video task I grepped (`videomme`, `longvt`, `streamingbench`, `motionbench`, `video_holmes`, ...) emits `{"type":"video","url": <local_path>}` — none produce remote URLs today, so I assume the use case is "let vLLM do server-side decode of a local file" and `file://` is just the scheme it needs. Is that right, or is there a task in mind that passes real HTTP URLs?

min1321:
Thanks for the careful review!

Agreed on both. Refactored as suggested: moved the video-URL branch into `protocol.to_openai_messages` behind a `pass_video_url: bool = False` flag, and `chat/openai.py` now just forwards the flag while keeping the `extra_body` composition. The local-image regression you spotted is a real concern — keeping the transform centralized avoids it. `to_qwen3_vl_openai_messages` can get the same hook in a follow-up.

On the use case — yes, exactly. We want vLLM to do server-side decoding of a **local** file so it can apply `media_io_kwargs.num_frames` and attach absolute-time signals (which the current client-side base64-frame extraction strips). No remote HTTP URLs in mind today; `file://` is just the scheme vLLM expects for local paths. Normalization in `to_openai_messages` is: if the path isn't already a URL scheme, prepend `file://<abspath>`.

kcz358 (approved):
LGTM, thanks for the contribution

PR #1377, opened 2026-07-06
fix(egotempo): parameterize the S3 tar-index fast path via env

Luodian:
The optional ego4d tar-extraction fast path in `egotempo/utils.py` hardcoded a specific object-store deployment: bucket name, Lance index URI, endpoint, and region, plus hand-parsed `~/.aws/credentials`. For anyone without that exact setup it was dead code behind a silent `except`.

This makes the fast path configuration-driven:

- `EGOTEMPO_LANCE_INDEX_URI` (new): Lance table indexing the ego4d tar shards (`member_path`, `tar_path`, `offset`, `size`). Unset -> fast path disabled, standard media fallback used (the default behavior for everyone today).
- Tar bucket derives from the index URI (shards live in the same bucket as the table).
- Credentials/region/endpoint resolve through the standard AWS chain (`AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `AWS_REGION` / `AWS_ENDPOINT_URL`, `~/.aws/*`) instead of string-splitting the credentials file.

No behavior change for regular users; deployments that used the previous hardcoded values set the two env vars instead.

Validation: `python -m py_compile` + ruff clean; with the env unset, `_get_s3_client()`/`_get_s3_index()` short-circuit to the fallback exactly as before.

### DIFF
--- a/lmms_eval/tasks/egotempo/utils.py
+++ b/lmms_eval/tasks/egotempo/utils.py
@@ -5,6 +5,7 @@ import tempfile
 from collections import defaultdict
 from pathlib import Path
 from typing import Any
+from urllib.parse import urlparse
 
 from lmms_eval.tasks._task_utils.media_resolver import resolve_media_reference
 from lmms_eval.utils import eval_logger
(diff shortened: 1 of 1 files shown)


Luodian:
Superseded: per data-hosting policy the public repo should not carry the S3/Lance fast path at all (even parameterized). Replacement PR removes all Lance-based data paths repo-wide, keeping HF/local resolution only.

PR #1378, opened 2026-07-06
fix: core evaluator/cache/utils correctness fixes

Luodian:
## Core evaluator / cache / utils correctness fixes

A cluster of standalone correctness fixes for the eval pipeline, response cache, and utils. No new dependencies; all changes are additive or in-place bug fixes.

### `lmms_eval/caching/cache.py`
- `save_to_cache` accumulated per-item results into `serializable_item` and appended to `serializable_obj`. Before, the loop mutated `subitem.arguments` but never appended anything, so the cache always pickled an **empty list**.
- Reviewer caveat: the request cache key does not fingerprint dataset contents, so with `--cache_requests` (opt-in, default off) a dataset changed in place under an unchanged task config will reuse stale requests. That key design predates this PR — flagging it because the cache now actually works and the caveat becomes reachable.

### `lmms_eval/caching/response_cache.py`
- Task fingerprinting now recurses into nested task groups (`_collect_fingerprints`), so subtasks get real config fingerprints instead of silently falling back to empty strings.

### `lmms_eval/evaluator.py`
- **Secret redaction**: sensitive keys are redacted from `results["config"]` before it is written out. Matching is suffix-based (`*api_key` / `*_token` / `*_secret` / `*password`, plus the exact-name set and `key=value` assignments inside `model_args` strings), so `openai_api_key=` and friends don't slip through.
- **Padding aliasing fix**: rank-padding built `[pad_source] * n`, aliasing a single `Instance` so all padding requests shared one `resps`/`token_counts` list. Now each padding slot is a distinct isolated clone (`_clone_padding_request` inside the loop).
- **Multi-rank auto-init (torchrun path only)**: on `distributed_executor_backend="torchrun"`, when `world_size > 1` and `torch.distributed` is uninitialized, auto-init a `gloo` process group so backends that never call `init_process_group` don't crash evaluator collectives — guarded on `MASTER_ADDR`/`MASTER_PORT` being present (clear warning + skip otherwise, instead of raising). The default `accelerate` path is untouched: the evaluator's `Accelerator()` fallback creates its own state, so GPU collectives are never silently downgraded onto a pre-created gloo group.
- **torchrun gather fix**: `all_gather_into_tensor` now uses a 1-D input tensor and a `world_size`-sized buffer.
- **Defensive skip**: docs that land in a rank's shard with no requests are skipped instead of crashing at `requests[0].doc`.

### `lmms_eval/utils.py`
- Robust `get_git_commit_hash` / `get_git_branch_name`: anchored at the repo root (works regardless of CWD), with env fallbacks (`LMMS_EVAL_GIT_COMMIT` / `GIT_COMMIT` / `GITHUB_SHA`, `LMMS_EVAL_GIT_BRANCH` / `GIT_BRANCH` / `BRANCH_NAME` / `GITHUB_REF_NAME`) and detached-HEAD handling.
- Local dataset builders (`csv`/`json`/`parquet`/`text`/`pandas`/`arrow`) now resolve relative `dataset_kwargs.data_files` paths to absolute against the YAML dir; HF-hub dataset paths are left untouched.
- py3.13 fix: `re.split(..., maxsplit=maxsplit)` (positional `maxsplit` is deprecated in 3.13).

### Validation
- `py_compile` clean on all four files; pre-commit black+isort clean; `ruff` introduces no new findings.
- Ported helpers unit-exercised: git resolution + env fallbacks + detached HEAD, `data_files` resolution, secret redaction (prefixed keys + string-form assignments), and padding-clone isolation (N distinct objects for N slots).

### Review pass (2026-07-06)
Self-review follow-up commit: gated the gloo auto-init to the torchrun path with the MASTER_ADDR/PORT guard (see above), moved the padding clone inside the loop so slots are truly isolated, switched redaction to suffix matching, and dropped an unused `build_eval_output_dir` helper that had no callers.

### DIFF
--- a/lmms_eval/caching/cache.py
+++ b/lmms_eval/caching/cache.py
@@ -44,10 +44,13 @@ def save_to_cache(file_name, obj):
     serializable_obj = []
 
     for item in obj:
+        serializable_item = []
         for subitem in item:
             if hasattr(subitem, "arguments"):  # we need to handle the arguments specially since doc_to_visual is callable method and not serializable
                 serializable_arguments = tuple(arg if not callable(arg) else None for arg in subitem.arguments)
                 subitem.arguments = serializable_arguments
+            serializable_item.append(subitem)
+        serializable_obj.append(serializable_item)
 
     eval_logger.debug(f"Saving {file_path} to cache...")
     try:
--- a/lmms_eval/caching/response_cache.py
+++ b/lmms_eval/caching/response_cache.py
@@ -407,14 +407,21 @@ class ResponseCache:
             cache_root = os.path.dirname(os.path.abspath(cache_root))
         cache_root = os.path.abspath(cache_root)
 
-        # Task fingerprints
+        # Task fingerprints — recursively process nested groups so subtasks
+        # get proper fingerprints instead of falling back to empty strings.
         task_fingerprints = {}
         if task_dict:
-            for tname, tobj in task_dict.items():
-                if hasattr(tobj, "dump_config"):
-                    cfg_str = json.dumps(tobj.dump_config(), sort_keys=True, default=str)
-                    cfg_str = _FUNC_ADDR_RE.sub(">", cfg_str)
-                    task_fingerprints[tname] = hash_string(cfg_str)[:16]
+
+            def _collect_fingerprints(d):
+                for tname, tobj in d.items():
+                    if hasattr(tobj, "dump_config"):
+                        cfg_str = json.dumps(tobj.dump_config(), sort_keys=True, default=str)
+                        cfg_str = _FUNC_ADDR_RE.sub(">", cfg_str)
+                        task_fingerprints[tname] = hash_string(cfg_str)[:16]
+                    elif isinstance(tobj, dict):
+                        _collect_fingerprints(tobj)
+
+            _collect_fingerprints(task_dict)
 
         # Model fingerprint
         if isinstance(model_args, dict):
--- a/lmms_eval/evaluator.py
+++ b/lmms_eval/evaluator.py
@@ -81,13 +81,20 @@ _SENSITIVE_CONFIG_KEYS = {
     "huggingfacehub_api_token",
     "token",
 }
-_SECRET_ASSIGNMENT_RE = re.compile(r"(?i)(^|[,\s])(api_key|client_secret|hf_token|huggingface_hub_token|huggingfacehub_api_token|token)=([^,\s]+)")
+# Suffixes also catch prefixed variants: openai_api_key, github_token, my_secret, db_password.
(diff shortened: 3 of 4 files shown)


Luodian:
@kcz358 @pufanyi This batch of 12 PRs (#1378–#1389) upstreams fixes and features from our fork — all now ready for review. Every PR went through a full review pass today; findings were fixed on the branches and each PR body has a "Review pass (2026-07-06)" section summarizing what changed. A map to make this reviewable:

**Suggested merge order** — plain numeric order (#1378 → #1389) avoids all conflicts. The actual constraints:
- #1378 before #1380 (both touch `evaluator.py`)
- #1379 before #1388 (stacked: #1388 = 10 task migrations on top of the verifiers module; once #1379 lands, #1388 gets rebased so the duplicated base commits drop out)
- #1382 before #1385 (both touch `pyproject.toml`) and before #1389 (both register models; #1389 inlines a minimal `ParallelPlan` that switches to #1382's `model_utils/parallel.py` — one-line dedupe at merge)
- #1386 before #1388 (both touch `tasks/videoevalpro/utils.py`; trivial re-merge either way)

**What's in the batch:**
| PR | Area |
|---|---|
| #1378 | core evaluator / cache / utils correctness fixes |
| #1379 | new `verifiers` module (pluggable per-sample verification) |
| #1380 | opt-in strip-thinking filter (`TaskConfig.auto_strip_thinking`) |
| #1381 | llm_judge OpenAI providers: base URL, reasoning-model params, PIL images |
| #1382 | gemini → google-genai SDK migration (video/audio upload, concurrency) |
| #1383 | HTTP eval server job-scheduler fixes |
| #1384 | remove Lance-based data paths (public repo = HF/local data only) |
| #1385 | SwanLab logger (`--swanlab_args`), wandb alternative |
| #1386 | utils fixes across 7 benchmarks |
| #1387 | IntPhys2 + VANTAGE-Bench tasks (PhysBench pulled — see PR for why) |
| #1388 | migrate 10 benchmarks from llm_judge to verifiers (stacked on #1379) |
| #1389 | 8 diffusers/API world-model generation backends |

**CI note — pre-existing, not from this batch:** `compare-task-input-boundary` is currently red on every task-touching PR in the repo (including merged #1374–#1376). Root cause: `lmms_eval/tasks/mmmu/utils.py` builds the LLM-judge client at import time (module-level `server = get_server(...)`), and the CI runner has no `OPENAI_API_KEY`, so the snapshot tool crashes while importing `mmmu_val` from its spec — before any comparison happens. Same reason `--tasks mmmu_val` fails for anyone without an OpenAI key. Happy to send a small follow-up making that init lazy, which would turn the check green repo-wide.

Ping me here or on the individual PRs with anything.

kcz358 (approved):

PR #1386, opened 2026-07-06
fix(tasks): utils improvements across 7 benchmarks

Luodian:
## Summary

Generic, self-contained improvements to seven evaluation task utilities. No new dependencies; all data still loads from the existing public Hugging Face datasets.

## Changes

- **egoschema**: Fix the `HF_HOME` cache path (`~/.cache/huggingface`, not `.../hub`). Replace the unbounded submission retry loop with a bounded, backed-off POST to EgoSchema's public validation server that saves the submission first and fails loudly on persistent failure instead of spinning forever or recording a bogus 0. Return the parsed accuracy from the MC / MC_PPL aggregators. Add deterministic `generation_kwargs`. README now spells out which tasks score offline (the N=500 subset) vs. which are submission-only (the full set) — and the offline subset tasks (`egoschema_subset`, `egoschema_subset_mcppl`) no longer declare the server `submission` metric at all: they have real answers and score locally, so an offline/no-egress node can't crash on the (now loud) server POST. The full-set tasks keep server submission.
- **erqa**: Use the shared `extract_mcq_answer` helper, add a `doc_to_messages` builder that interleaves images before the question text, and remove the dead local extractor / unused template load. Adds unit tests.
- **vsibench**: Robust numeric-answer extraction for NA-type questions — handles bare digits, digits embedded in a sentence, and English number words (e.g. `twenty-one`, `three hundred and twenty-one`), with the first number token winning.
- **vsisuper**: Harden the count_streaming metrics — `_abs_dist_norm` now coerces to float and guards divide-by-zero, and `_parse_prediction_list` tolerates malformed / non-list JSON. **Scoring note**: this corrects a bug where malformed predictions and zero-valued targets scored a *perfect* MRA (the old `except: return 0.0` fed the "distance" side); they now score worst-case, so affected historical count_streaming numbers will drop.
- **mmsi_video**: Add a single-visual-input subset (`mmsi_video_single_input`) matching Cambrian-P's reported MMSI-Video protocol, and drop unused imports.
- **tomato**: Make the OpenAI import lazy (only needed for the disabled GPT parser) and stop hardcoding a 16-frame assumption in the prompt so the model's own `num_frames` governs the frame budget.
- **videoevalpro**: Guard the score aggregator against an unexpected `task_type` (previously a `KeyError`) and use identity comparison for the judge flag.

## Validation

- `py_compile`, `ruff check`, `black --check` (line-length 240), and `isort --profile black --check` all pass on the changed files.
- YAML task configs verified; all referenced `!function` utilities exist.
- New `erqa/tests/test_erqa_utils.py` covers answer extraction, message interleaving, and aggregate accuracy.
- Offline verification of the egoschema metric change: `TaskManager` resolves `egoschema_subset`/`egoschema_subset_mcppl` with `metric_list=['score']` and the full `egoschema`/`egoschema_mcppl` still with `['submission']`, no network access.

### Review pass (2026-07-06)
Self-review follow-up commit removes the `submission` metric from the two offline subset yamls (see egoschema bullet) — previously the README recommended the subset for offline eval while its submission metric still POSTed to the validation server and, after this PR's own retry hardening, would have raised on a no-egress node before the offline `score` was emitted.

### DIFF
--- a/lmms_eval/tasks/erqa/tests/test_erqa_utils.py
+++ b/lmms_eval/tasks/erqa/tests/test_erqa_utils.py
@@ -1,0 +1,32 @@
+from PIL import Image
+
+from lmms_eval.tasks.erqa import utils
+
+
+def test_erqa_process_results_extracts_final_choice():
+    doc = {"answer": "B", "question_id": "ERQA_1", "question_type": "Action Reasoning"}
+
+    result = utils.erqa_process_results(doc, ["I considered A, but the answer is B."])
+
+    assert result["erqa_acc"]["is_correct"] is True
+    assert result["erqa_acc"]["id"] == "ERQA_1"
+
+
+def test_erqa_doc_to_messages_interleaves_images_before_text():
+    image = Image.new("RGB", (2, 2), color="white")
+    doc = {"images": [image], "question": "Choices: A. yes B. no"}
+
+    messages = utils.erqa_doc_to_messages(doc)
+
+    assert messages[0]["role"] == "user"
+    assert messages[0]["content"][0]["type"] == "image"
+    assert messages[0]["content"][1] == {"type": "text", "text": doc["question"]}
+
+
+def test_erqa_aggregate_results_returns_accuracy():
+    results = [
+        {"sub_task": "Action Reasoning", "is_correct": True},
+        {"sub_task": "Action Reasoning", "is_correct": False},
+    ]
+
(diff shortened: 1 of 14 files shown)


lmms_eval/tasks/erqa/tests/test_erqa_utils.py
kcz358, line 1: This file and tests folder can be removed

PR #1391, opened 2026-07-06
fix(distributed): call destroy_process_group() on exit to prevent NCC…

### LINKED ISSUE #1240: [Bug] NCCL warning on error path: destroy_process_group() not called
## Summary
In distributed runs (`accelerate launch ...` / multi-GPU), when evaluation hits an exception, process shutdown can emit:

```text
[W ... ProcessGroupNCCL.cpp:1479] Warning: WARNING: destroy_process_group() was not called before program exit, which can leak resources.
```

This looks like missing explicit distributed teardown on error paths.

Verified against `upstream/main` commit `ef732077`.

## Why this happens
`lmms_eval/__main__.py` initializes an `Accelerator` (which sets up distributed state) before entering the evaluation loop:

```python
# lmms_eval/__main__.py
if torch.distributed.is_available() and torch.distributed.is_initialized():
    accelerator = None
else:
    kwargs_handler = InitProcessGroupKwargs(timeout=datetime.timedelta(seconds=60000))
    accelerator = Accelerator(kwargs_handlers=[kwargs_handler])
```

Then exceptions are caught and logged, but there is no `finally` teardown:

```python
# lmms_eval/__main__.py
for args in args_list:
    try:
        results, samples = cli_evaluate_single(args)
        ...
    except Exception as e:
        ...
        results_list.append(None)
```

`lmms_eval/evaluator.py` has synchronization (`wait_for_everyone` / `barrier`) but no explicit `destroy_process_group()` either.

## Repro (one simple way)
Any multi-GPU run that throws after distributed init can trigger it, for example using intentionally invalid model args:

```bash
accelerate launch --num_processes=2 -m lmms_eval \
  --model qwen3_vl \
  --model_args pretrained=Qwen/THIS_MODEL_DOES_NOT_EXIST \
  --tasks mme \
  --batch_size 1
```

Expected failure is the invalid model load; after that, process exit may print the NCCL warning above.

## Impact
- Noisy logs and confusing secondary warning after the primary error
- Potential resource leakage in long-running orchestration environments

## Possible fix direction
Add explicit distributed cleanup in a top-level `finally` block (guarded by `torch.distributed.is_available()` and `torch.distributed.is_initialized()`), e.g. call `torch.distributed.destroy_process_group()` once per process before exit.

mandanajignesh-byte:
…L resource leak

In distributed runs (accelerate launch / multi-GPU), when evaluation hits an exception, process shutdown emits:

  WARNING: destroy_process_group() was not called before program exit,
  which can leak resources.

Root cause: Accelerator registers no atexit handler, so distributed teardown never happens on error paths.

Fix: wrap the evaluation loop and result printing in a try/finally block. The finally block calls accelerator.state.destroy_process_group() when Accelerator is active, or torch.distributed.destroy_process_group() directly when running in a pre-initialized distributed context. wandb logging runs inside try before finally to preserve correct ordering.

Fixes #1240

## Summary

- 
- 
- 

## In scope

- 

## Out of scope

- 

## Validation

- 

## Risk / Compatibility

- 

## Type of Change
- [ ] New feature
- [ ] New benchmark/task
- [ ] New model integration
- [ ] Breaking change
- [ ] Documentation update


pufanyi:
Hello! Thanks for your contribution! I have resolved the conflict and will merge it to main.

pufanyi (approved):

PR #1403, opened 2026-07-21
[Model] Add Neo-OV model support and SI benchmark configs

oscarqjh:
## Summary

This PR adds `neo_ov` as a new lmms-eval chat model backend and aligns benchmark-specific prompting / generation settings for Neo-OV evaluation on 8 spatial intelligence benchmarks.

  Affected benchmarks:

  - MindCube Tiny
  - MMSI-Bench
  - ViewSpatial
  - SiteBench
  - BLINK
  - 3DSRBench
  - EmbSpatial
  - VSI-Bench

  ## Changes

  - Add the `neo_ov` model wrapper and registry entry.
  - Add Neo-OV-specific task YAML settings via `lmms_eval_specific_kwargs` and `model_specific_generation_kwargs`.
  - Keep prompt construction in task-level `doc_to_messages` utilities so the model wrapper remains generic.
  - Align Neo-OV settings with the corresponding SI benchmark evaluation behavior used in [original repo](https://github.com/EvolvingLMMs-Lab/NEO/tree/main/VLMEvalKit_ov) where applicable.

## Validation

  - Reproduced the Neo-OV spatial intelligence benchmark behavior/results against the official NEO VLMEvalKit-ov evaluation setup:
  https://github.com/EvolvingLMMs-Lab/NEO/tree/main/VLMEvalKit_ov
  - Ran `black` and `isort` pre-commit hooks on changed Python files.

## Example
Evaluation of Paranioar/NEO1_5-2B-SFT on vsibench
image

### DIFF
--- a/lmms_eval/models/__init__.py
+++ b/lmms_eval/models/__init__.py
@@ -127,6 +127,7 @@ AVAILABLE_CHAT_TEMPLATE_MODELS = {
     "internvl_hf": "InternVLHf",
     "llava_hf": "LlavaHf",
     "nanovlm": "NanoVLM",
+    "neo_ov": "NeoOV",
     "phi4_multimodal": "Phi4",
     "qwen3_vl": "Qwen3_VL",
     "qwen3_5": "Qwen3_5",
--- a/lmms_eval/models/chat/neo_ov.py
+++ b/lmms_eval/models/chat/neo_ov.py
@@ -113,11 +113,7 @@ def _preprocess_pixel_values(pixel_values: torch.Tensor, patch_size: int = 16) -
     c, h, w = pixel_values.shape
     grid_h = h // patch_size
     grid_w = w // patch_size
-    flat_pixel_values = (
-        pixel_values.view(c, grid_h, patch_size, grid_w, patch_size)
-        .permute(1, 3, 0, 2, 4)
-        .reshape(grid_h * grid_w, c * patch_size**2)
-    )
+    flat_pixel_values = pixel_values.view(c, grid_h, patch_size, grid_w, patch_size).permute(1, 3, 0, 2, 4).reshape(grid_h * grid_w, c * patch_size**2)
     return flat_pixel_values, torch.tensor([[grid_h, grid_w]], device=pixel_values.device)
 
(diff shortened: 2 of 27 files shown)


oscarqjh:
@pufanyi @kcz358

kcz358 (commented):
Hi, most of the parts LGTM for me. For the model specific generation kwargs, I saw that all the parameters also exists in the model args, so I am wondering if this is better to pass in through model args with runtime script instead of directly adding in the yaml? Thanks

oscarqjh:
Hi @kcz358 , thanks for the review!
My rationale for pinning the model-specific generation kwargs in the YAML is to improve convenience and reproducibility for researchers. Some models, like Neo-OV, heavily tune their generation kwargs to achieve optimal performance on specific benchmarks. For example, here is how it is handled in VLMEvalKit's configs: https://github.com/EvolvingLMMs-Lab/NEO/blob/87ab2803dd288e3265ba78fd61d250f64bcf1bdc/VLMEvalKit_ov/vlmeval/config.py#L1757-L1806
image


Baking these into the YAML saves users from having to dig through Neo-OV's official repo to hunt down the correct configurations.

That said, if you prefer to keep the YAML files clean and agnostic of model-specific kwargs, I am happy to remove them. Let me know what you think works best for the repo!

kcz358:
Hi, sorry for the late reply. I am okay for putting that into the yaml since it is harmless for the yaml. It seems that the neo itself has many different set of parameters for different size of model? Would that mean people still needs to dig into the repo for the parameters. If that is true, I think maybe you can add the default params in the model file itself as a default? Similar to the vlmevalkit that if using NEO 2b image then would be this parameters.

kcz358:
Hi @oscarqjh, I'm planning to merge this now. If there are any other model-specific parameters or settings that users should be aware of, please feel free to document them later in the model README, or add a section in the model repository describing the detailed lmms-eval configuration.

oscarqjh:
@kcz358 Okay, thank you. Sorry, I was not free these few weeks. Will see if I can afd any further clarification in the readme later.

PR #1418, opened 2026-07-31
feat: add agentic game-loop evaluation (generate_until_game)

Luodian:
## Summary
- Adds `generate_until_game` for evaluating a model as a policy in an interactive environment, with ViZDoom as the first task.
- Task YAML owns `game_env` and `model_specific_parsers`; the CLI owns the model server and rollout concurrency.
- Parsers are ordinary task-local `(value, ParserContext) -> value` functions. A pipeline may transform any value into any other value; the framework only validates `AgentInput` and `ParsedAction` at the model-server and environment boundaries.
- `model_specific_parsers` selects by the actual model identity reported by the server, with exact keys and case-insensitive globs such as `*Qwen*`; specific entries inherit omitted stages from `default`.
- Episodes run as synchronous loops parallelized by a thread pool; OpenAI-compatible endpoints batch concurrent requests server-side.

## In scope
- `lmms_eval/agentic/`: shared open-ended types, `ParserContext`, parser-pipeline selection/composition, `EnvManager`, episode loop, trace, artifacts, runner, and focused model-server modules (`servers/{base,debug,openai}.py`).
- `lmms_eval/tasks/vizdoom_agentic/`: task-local ViZDoom environment and parser functions. The Qwen normalization function lives here and is composed before the ViZDoom action parser through the task YAML.
- `ModelServer.get_model_name()`: optional model identity for task parser selection; OpenAI uses its `model=...` argument and debug reports `debug`.
- Removes the framework parser ABC/registry and `--agentic_output_parser`; task YAML is the single owner of model/task parsing behavior.
- Minimal core hooks for `output_type: generate_until_game`, unit tests, and docs.

## Parser contract
```yaml
model_specific_parsers:
  default:
    observation: !function parsers.vizdoom_observation_parser
    action: !function parsers.vizdoom_action_parser
  "*Qwen*":
    action:
      - !function parsers.vizdoom_qwen_output_parser
      - !function parsers.vizdoom_action_parser
```

Exact model keys win over globs. Each stage accepts either one callable or a callable list; model-specific stages merge over `default`.

## Out of scope
- No changes to existing output types, tasks, or model backends.
- Multi-agent episodes, async rollout scheduling, additional environments, and remote environment serving.

## Validation
- `.venv/bin/python -m pytest test/agentic test/eval/test_construct_requests.py -q` | `71 passed`
- Isolated-cache TaskManager load: 120 ViZDoom rows; `Qwen/Qwen3.6-27B` selects `[vizdoom_qwen_output_parser, vizdoom_action_parser]`.
- Real ViZDoom debug-server rollout: 60 steps, `vizdoom_invalid_actions=0`, completed end to end.
- black/isort pass; GitHub lint and task-input boundary checks pass on `12452a25`.

## Risk / Compatibility
- Additive relative to `main`: the output type and CLI flags do not affect existing tasks.
- This supersedes the parser API from earlier revisions of this unmerged PR: parser classes/registries and `--agentic_output_parser` are intentionally removed.
- ViZDoom remains an optional dependency loaded inside the task; core `lmms_eval.agentic` imports stay lightweight.

## Type of Change
- [x] New feature
- [x] New benchmark/task
- [ ] New model integration
- [ ] Breaking change
- [x] Documentation update

## Credits
The agentic game-loop direction builds on @pufanyi's ViZDoom agent design in #1368 — the ViZDoom task here carries that groundwork onto the standard interface. Thanks Fanyi!


pufanyi:
Parser design update (`12452a25`):

- removed the framework-level parser ABCs, implementations, and parser registry
- moved every concrete parser to its task directory; ViZDoom parsing remains in `tasks/vizdoom_agentic/parsers.py`
- changed the runtime contract to task-local `(value, ParserContext) -> value` functions, composed as `Any -> Any` pipelines
- added task YAML `model_specific_parsers` selection by the serving model identity, supporting exact keys and case-insensitive globs such as `*Qwen*`
- moved Qwen normalization into the ViZDoom task and configured it as `[vizdoom_qwen_output_parser, vizdoom_action_parser]`
- removed `--agentic_output_parser`; model/task parsing now has one owner in task YAML
- kept the focused `servers/{base,debug,openai}.py` package split

The stacked child #1419 merged this parent in `08c1e96f`, so both branches use the same parser protocol and the child diff does not reintroduce framework parser files.

Validation: black/isort pass; `71 passed`; TaskManager loads all 120 ViZDoom rows and selects the Qwen pipeline for `Qwen/Qwen3.6-27B`; a real ViZDoom debug rollout completed 60 steps with 0 invalid actions; both GitHub checks pass.

kcz358 (commented):
**Summary:** The goal is right and the result works — ViZDoom does run end-to-end, and the multi-tier action parser encodes real debugging experience. But the architecture is sized for a problem we don't have yet, and ~80% of what it builds already exists in `generate_until_agentic`. I'd like to see this split up and rebuilt on the existing path.

---

### 1. The abstraction layer is far larger than the set of things it abstracts — and it doesn't actually constrain anything

Current implementation counts:

| Abstraction | Implementations |
|---|---|
| `EnvManager` | 1 |
| `ModelServer` | 2 (one is a 20-line debug stub) |
| `REGISTRY` kinds | 1 |
| spec forms supported by `resolve()` | 5 declared, 2 used |
| `model_specific_parsers` glob patterns | 1 |
| multi-agent | 0 |

More importantly, these types don't buy type safety. `GameAction` is:

```python
type: str
data: Any = None
```

That has degenerated to `(str, Any)`. The proof is `env.py:_apply_action_data`, which has to accept five different payload shapes (`list[float]` full vector / `list[str]` button names / `dict{"buttons"}` / `dict{"values"}` / `dict{"button_values"}`) precisely because the type says nothing about what `data` is. **We pay the indirection cost and get no safety in return.**

Same pattern elsewhere:

- Parser pipelines are `Callable[[Any, ParserContext], Any]`. Types are checked only at the two ends of the pipeline via `isinstance`; the middle is unconstrained. That's a convention plus a runtime assertion, not an abstraction.
- Multi-agent is fully plumbed at the type level (`GameAction.agent_id`, `step(dict[str, GameAction])`, `reward: float | dict[str, float]`, `ParserContext.agent_id`), but `env.py:step` opens with `action = next(iter(action.values()))`, discarding every agent but the first, and `episode.py` hardcodes `agent_id="agent"`.
- `ContentBlock`'s docstring claims support for JEPA-style `tensor`/`embedding`/`latent`/`logits`. No implementation, no test.

**Suggestion:** don't model input/output as a type hierarchy. A general kwargs/fields channel is enough and won't need escape hatches later.

The same applies to `VizDoomEnvManager.__init__`: 60+ kwargs mirroring `DoomGame.set_*`, each repeated three times (signature → `self.config` → `direct_settings`), roughly 200 lines. The class already has a generic channel that subsumes all of it:

```python
settings: dict[str, Any] | None = None,
**extra_settings: Any,
...
for key, value in self.settings.items():
    self._apply_setting(game, key, value)   # getattr(game, f"set_{key}")(value)
```

And `utils.vizdoom_env_manager`, the only call site, hardcodes nearly all 60 of them. We parameterized 60 dimensions and then froze every one.

---

### 2. Parsers and prompts should be task-level, not a pipeline abstraction

Commit `12452a25` moved the parser implementations into the task directory — right direction. But the framework side stayed: the same commit added `pipelines.py`, the `Parser` type, `ParserContext`, and the `model_specific_parsers` YAML key. So the concrete code lives in the task while the machinery dispatching it lives in the framework, which means every future agentic task inherits a parser pipeline shape whether it wants one or not.

I don't think we need a framework concept here at all. The task can parse `previous_output[-1]` however it likes, as ordinary code — `vending_bench2` already does exactly this:

```python
model_response = previous_output[-1] if previous_output else ""
tool_payloads = _extract_tag_payload(TOOL_CALL_PATTERN, model_response)
```

If another task later wants ViZDoom's tool-call extraction or button decoding, it can just `from lmms_eval.tasks.vizdoom_agentic.utils import ...`. That's a cheap, explicit dependency, and it stays honest about where the code actually belongs.

`model_specific_parsers` is the part I'd most like to drop. Selecting parsers by glob-matching the model name means every agentic task has to know the output quirks of every model family, and adding a model means editing task YAMLs. A YAML variant, or a flag in `lmms_eval_specific_kwargs`, is simpler and reproducible. The `<think>`-stripping half of it also overlaps with the existing `reasoning_tags` mechanism.

---

### 3. Occam's razor: this should be an extension of `generate_until_agentic`, not a parallel subsystem

`generate_until_agentic` already provides most of what's needed:

- the loop lives in `evaluator.py:678-822`, so models only need `generate_until`
- `max_agentic_steps` bound at `evaluator.py:700`
- per-round visuals already swappable at `evaluator.py:787-788`
- `previous_round_info` is an arbitrary Python object passed through untouched, so it can carry a live environment handle

And there's already a stateful-simulator precedent — `vending_bench2`:

```python
def vending_doc_to_text(doc, ..., previous_output=None, round_idx=None, previous_round_info=None):
    state_info = previous_round_info or {"state": copy.deepcopy(doc["initial_state"]), ...}
    tool_result = _apply_tool(tool_name, arguments, state)      # advances the env in place
    next_prompt = _build_agent_prompt(doc, state, tool_result=...)
    return [], next_prompt, False, previous_output, state_info
```

That *is* `env.step()`. The only difference for ViZDoom is that it needs to return image frames rather than text.

The blocking gap is small and specific. `ConfigurableMessagesTask.doc_to_messages` already has the multi-round signature (`task.py:1822`):

```python
def doc_to_messages(self, doc, *, round_idx=None, previous_output=None, previous_round_info=None):
```

but the agentic branch never wires it up (`task.py:1844-1852` forces `doc_to_visual` + `doc_to_text`), and `evaluator.py:721-734` hardcodes a single fresh user message per round with no assistant history. Text-based tasks work around this by serializing state into the prompt; image frames can't be serialized that way.

**Estimated fix: ~30 lines of framework change**, versus ~3800 here. That gap alone is worth re-staging for.

---

### 4. On going through an OpenAI-compatible server

I understand the intent — standardizing on vLLM/SGLang-style OAI endpoints so we don't maintain N model wrappers. No objection to that as a serving strategy.

But I'd put the abstraction boundary in a different place. Right now `runner.py` only uses `lm` to fetch the doc (`lm.task_dict[task][split][doc_id]`) and resolves the model separately from `--agentic_model_server`, so the contract is "must be an OpenAI-compatible HTTP server." I think the contract should be at the `generate` level instead — something `generate_until`-shaped — so that if we ever swap SDKs or need a non-HTTP backend, we're implementing a `generate` method rather than emulating an HTTP protocol. That also keeps `--model` / `--model_args` and the existing post-processing path meaningful.

Concretely: keep the loop calling `lm.generate_until(...)`, and let "OAI endpoint" be one wrapper behind that, which is what it already is today.

---

### 5. Blocking bugs, independent of the architecture discussion

**a. Race in `artifacts.py:_new_artifact_dir`** — checks `while artifact_dir.exists()` and then calls `mkdir(exist_ok=False)`. Under `--agentic_max_parallel_rollouts > 1`, two threads can select the same path and the second one dies with `FileExistsError`, failing the episode. Move the collision handling inside a `try/except FileExistsError` retry.

**b. No environment lifecycle hook on the loop's exit paths** — this applies to either architecture. The agentic loop has four exits (terminal signal / step budget exhausted / malformed payload / exception). `DoomGame` is a subprocess; leaks accumulate until the box is saturated. Needs a `try/finally` plus a convention for closing whatever is held in `previous_round_info`.

**c. No per-episode error isolation** — `runner.py` calls `future.result()` directly, so one failed rollout aborts the entire evaluation. Env crashes and endpoint timeouts are routine in long rollouts; a single failure should be recorded as a failed sample.

**d. `agentic_trace_mode: full` will try to `json.dumps` `previous_round_info`** (`evaluator.py:769-773, 814-820`). A live env handle in there blows up. There's a `except (TypeError, json.JSONDecodeError): pass` catching it, which means the trace is silently dropped instead. Needs an explicit convention for what goes into the trace.

---

### 6. Minor

- `runner.py`: `max_steps = int(gen_kwargs.pop("max_game_steps", gen_kwargs.pop("max_agentic_steps", 32)))` — the inner `pop` always runs. Same pattern in `servers/openai.py:_build_openai_params` with `max_tokens`/`max_new_tokens`. Harmless today, wrong semantics.
- `EpisodeStep.output` is always identical to `raw_output` (`episode.py`), and `trace.py` serializes both. Either give `output` real meaning or drop it.
- `Instance.args[0]` (`ctx`) and `args[2]` (`doc_to_visual`) are unpacked as `_ctx` / `_doc_to_visual` and discarded; the runner re-fetches the doc from `lm.task_dict`. Two slots in the 9-tuple are dead, and the `doc_to_text` / `doc_to_visual` declared in `vizdoom.yaml` have no effect on the actual prompt.
- Loop parameters (`max_game_steps`, `game_seed`, `multiturn`, `history_turns`) ride inside `generation_kwargs` and then have to be filtered back out downstream via `_AGENTIC_ONLY_KEYS`. A field placed in the wrong container and dug out later.
- `_normalize_top_p` silently rewrites `top_p=0` to `1.0` with no log and no documentation.
- Artifacts are written unconditionally whenever `--output_path` is set — `summary.md` + `actions.jsonl` + `rollout.mp4` + per-step segment mp4s, at 4x upscale by default. At scale this is a lot of disk. Should be an explicit flag, not just env-var tuning.
- `.gitignore` adds `_vizdoom.ini`, but `utils.py` already writes the ini into a tempdir. Dead rule.

---

### What I'd keep

To be clear, this is a request to re-stage, not to discard. The following are real contributions and should survive:

| Keep | Why |
|---|---|
| `parsers.py` action decoding (5-tier fallback + natural-language rescue) | Pure functions, encodes actual debugging experience, portable to any architecture |
| `_skill_prompt_lines` prompt template | Same — things like "never combine MOVE_LEFT with MOVE_RIGHT" came from observed failures |
| `env.py` core (`reset` / `step` / `_state` / `_action_to_vector`) | ~150 lines after dropping the setter passthrough |
| `tics_per_action` + `frame_history` observation design | This is the one genuine evaluation-design decision in the PR ("the model sees the 5-frame clip produced by its last action, then picks the next one"). It deserves its own doc section instead of being buried among 60 kwargs. |
| `safe_data` | Useful as a standalone trace utility |

What I'd drop for now: the `lmms_eval/agentic/` package (11 dataclasses, `components.resolve` spec polymorphism, `REGISTRY`, `pipelines` glob dispatch, `servers`, `runner`, `episode`), and the ~200 lines of `DoomGame` setter passthrough in `env.py`.

---

### Suggested split

1. **`feat: wire doc_to_messages into generate_until_agentic`** — `task.py:1844-1852` + `evaluator.py:721-734`, plus the env lifecycle hook and a response-cache bypass for stateful tasks. ~30 lines of framework change. Reviewable and testable on its own, and it's a net win for the existing `vending_bench2` / `tau2_bench` tasks too, which currently also have no conversation history.
2. **`feat(vizdoom): add agentic ViZDoom task`** — env + parsers, ~400 lines, entirely inside `tasks/vizdoom_agentic/`.
3. **Follow-ups, independent:** trace/artifacts as a standalone utility extending the existing `agentic_trace_mode`; rollout parallelism as a thread pool around the existing request loop.

One practical note on (2): with `max_game_steps: 64` and a serial loop, each doc is 64 sequential round-trips. That's fine for a demo, but please cap the doc count in the first version — otherwise the serial path will look unusable for the wrong reason.

---

### On timing

The abstractions here (multi-agent, async envs, non-text policy models) are aimed at real future problems. But their shape can't be determined in advance — it only becomes visible after two or three real environments exist. Right now we have one env, one server, one task. I'd rather under-abstract now and refactor with evidence than lock in a shape that the next environment has to work around.

pufanyi:
> That has degenerated to `(str, Any)`.

I'm not sure if this is the right approach, but personally, I feel that because the type of model inputs and outputs are uncertain, using `Any` makes the model much easier to pick up without needing to prescribe too many rules. This kind of design makes it much simpler to add various parsers and models.

PR #1435, opened 2026-08-18
fix(blink): restrict answer extraction to valid choices and parse verbose responses

### LINKED ISSUE #1434: [Bug] BLINK answer extraction misparses verbose responses ("The correct answer is (B)" → "T"), penalizing small/verbose models
### Checklist

### Bug Description

`_extract_answer_letter()` in `lmms_eval/tasks/blink/utils.py` (introduced in #927) parses the model response with a start-anchored regex that is **not restricted to the valid option letters**:

```python
match = re.match(r"[\(\s]*([A-Z])[\)\.\s]*", text, flags=re.IGNORECASE)
```

Two consequences:

1. **Correct answers are graded as wrong.** A verbose response ending with `"... The correct answer is (B)."` is parsed as `T` (first letter of "The") and scored incorrect, even though the answer is trivially recoverable from the text.
2. **Spurious credit is possible.** A response starting with "Based on ..." parses as `B`; if `B` happens to be the ground truth, the sample is credited by accident. The parser can bias scores in **both** directions.

Since the extracted letter can be any capital letter rather than one of the doc's choices, this is not strict grading (which would return `""` → incorrect with a clear reason) — it is a parsing defect. Note that `lmms_eval/tasks/_task_utils/mcq_extract.py` (#1276, used by erqa/cgbench/lvbench/…) already implements robust extraction, but BLINK does not use it.

### Steps to Reproduce

```shell
python -m lmms_eval --model vllm_generate \
  --model_args model=Qwen/Qwen3-VL-2B-Instruct \
  --tasks blink --batch_size 64 --log_samples --output_path results/
```

Then inspect the sample logs: verbose responses whose final line is `The correct answer is (B).` with target `(B)` are scored 0 (`pred_parsed: "T"`).

Real sample from `blink_semantic_correspondence` (target `(B)`, scored 0):

```
The reference point 'REF' on the first image is located on the lower part of the
bird's leg ... Therefore, point B is the correct match for the reference point.

The correct answer is (B).
```

### Measured impact

We evaluated Qwen3-VL-2B/4B/8B-Instruct on `blink` (val, 1,901 docs, all 14 subtasks) and re-scored every sample offline with a robust extractor (letters restricted to each doc's choices; fallback: last `answer is (X)` / `(X)` occurrence in the full text):

| Base model | Official `blink` acc | Robust re-score | Δ | Letter-only compliance | Misgraded samples |
|---|---:|---:|---:|---:|---:|
| Qwen3-VL-2B | 43.29 | 51.71 | +8.42 | 45.1% | 160 / 1901 |
| Qwen3-VL-4B | 64.23 | 64.23 | 0 | 100.0% | 0 |
| Qwen3-VL-8B | 64.97 | 64.97 | 0 | 100.0% | 0 |

Observations:

- The 2B model's compliance with the prompt's "respond only with the letter" instruction collapses exactly on the hard subtasks (semantic/functional correspondence, jigsaw, forensic detection: 1–3% compliance), where it falls back to thinking out loud; official scores there drop **below random** (semantic_correspondence 1.4 vs 54.0 after re-scoring, functional_correspondence 0.0 vs 30.0, visual_correspondence 31.4 vs 59.3). On easy subtasks its compliance is 89–98%.
- Models ≥4B answer with a bare letter 100% of the time, so **fixing the parser is a no-op for compliant models** — it only removes misgrading. This also explains why the defect goes unnoticed with mainstream (≥4B) models.
- Practical consequence: in an SFT-vs-base comparison, official BLINK scoring showed the SFT model +5.1 over base; after robust re-scoring the base model is actually ahead (51.7 vs 48.9) — the parser artifact **inverted the conclusion**. Any BLINK evaluation of small or verbose/thinking-style models is affected.

### Environment

- OS: Ubuntu (Linux 5.4), Python 3.12
- lmms-eval: 0.7.2 (defect verified present on current upstream `main`, `lmms_eval/tasks/blink/utils.py`)
- vllm 0.19.1, transformers 5.15.0, torch 2.10.0+cu128, NVIDIA A800

### Suggested fix

At minimum, restrict the regex to the doc's valid option letters and return `""` when nothing matches (faithful strict grading). Preferably, reuse the existing `_task_utils/mcq_extract.py` or add a fallback cascade (terse leading letter → last `answer is ([A-D])` → last `\(([A-D])\)`). Happy to submit a PR with the fix and the re-scoring data above.

akawincent:
#### The bug

BLINK grades a response by grabbing the **first capital letter** it sees. That's fine for `B` or `(B) Point B` — but the moment a model explains itself, grading falls apart. A real response from Qwen3-VL-2B:

> The reference point 'REF' is located on the lower part of the bird's leg ... **The correct answer is (B).**

Ground truth: `(B)`. Old parser: extracts `T` — from the word *"The"* — and marks it **wrong**. The mirror image also exists: a response opening with *"Based on..."* parses as `B`, which silently becomes a lucky *hit* whenever B is the ground truth. Models are being graded on their writing style, not their perception. Full measurements in #1434.

#### The fix

 Read the response the way a human grader would:

1. If the reply is just a letter (`B`, `(B).`) → take it.
2. Otherwise find where the model *declares* its answer — “the answer is (B)” — and take the **last** such declaration (final answers come last in chain-of-thought).
3. Failing that, take the last parenthesized option like `(C)` mentioned anywhere.
4. As a last resort, keep the old leading-letter behavior.

One rule guards every step: **only letters that are actual choices for this question count** (derived from `doc["choices"]`), so *"The..."* can never again produce a phantom `T`. If nothing matches, return `""` and mark it wrong — that's what strict grading should look like.

Fixes #1434.

## In scope

- `lmms_eval/tasks/blink/utils.py` only: `_extract_answer_letter()` + its call site in `blink_process_results()`.
- (Second commit is unrelated lint debt on `main` — see PR comment below.)

## Out of scope

- Prompts, YAMLs, `doc_to_text` / `doc_to_messages`, aggregation, other tasks — untouched. The `compare-task-input-boundary` check confirms task inputs are byte-identical.

## Validation

- **Does it fix the problem?** Re-scored the logged `blink` val responses of Qwen3-VL-2B-Instruct | `N=1901` | acc `43.29 → 51.71`, recovering 160 answers that were correct but misparsed. Subtasks that sat *below random* under the old parser (Semantic_Correspondence `1.4%`, Functional_Correspondence `0.0%`) return to sane levels (`54.0` / `30.0`) | result: pass
- **Does it disturb well-behaved models?** Qwen3-VL-4B / 8B answer with a bare letter on 100% of val docs | `N=1901` each | scores bit-identical before vs after: `64.23` / `64.97` | result: pass
- **Are the old formats still parsed the same?** All six docstring formats (`A answer1`, `A) answer2`, `(B) answer`, `C`, `(C)`, `A.`) → unchanged letters | result: pass

## Risk / Compatibility

- Scores only move for responses the old regex misread (verbose / thinking-style outputs, mostly small models). Terse letter answers take the exact same code path, so existing numbers for compliant models don't shift.
- A response that never states a valid option letter still scores 0 — strictness is preserved, only the misreads are gone.

## Type of Change

- [ ] New feature
- [ ] New benchmark/task
- [ ] New model integration
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/lmms_eval/models/chat/aero_realtime_vllm.py
+++ b/lmms_eval/models/chat/aero_realtime_vllm.py
@@ -88,8 +88,8 @@ SAMPLE_RATE = 16000
 
 @dataclass
 class RealtimeSample:
-    audio: np.ndarray            # mono float32 @ 16 kHz
-    video: np.ndarray            # (T, H, W, 3) uint8
+    audio: np.ndarray  # mono float32 @ 16 kHz
+    video: np.ndarray  # (T, H, W, 3) uint8
     video_metadata: object
     sample_fps: float
 
@@ -339,10 +339,7 @@ class AeroRealtimeVLLM(lmms):
         # vs. ``aero_realtime_chat`` which is strict, to match how other vLLM
         # wrappers behave.
         if AsyncOmni is None or AeroRealtimeForConditionalGeneration is None:
-            raise ImportError(
-                "vllm-omni is required for aero_realtime_vllm_chat. "
-                "Install vllm-omni and ensure it is importable."
-            )
+            raise ImportError("vllm-omni is required for aero_realtime_vllm_chat. " "Install vllm-omni and ensure it is importable.")
         if fetch_video is None:
             raise ImportError("qwen_vl_utils is required (`pip install qwen-vl-utils`)")
         if librosa is None:
@@ -370,9 +367,7 @@ class AeroRealtimeVLLM(lmms):
             omni_kwargs["stage_init_timeout"] = int(stage_init_timeout)
         if tensor_parallel_size is not None:
             omni_kwargs["tensor_parallel_size"] = tensor_parallel_size
-            omni_kwargs["stage_0_devices"] = stage_0_devices or ",".join(
-                str(i) for i in range(tensor_parallel_size)
-            )
+            omni_kwargs["stage_0_devices"] = stage_0_devices or ",".join(str(i) for i in range(tensor_parallel_size))
         elif stage_0_devices is not None:
             omni_kwargs["stage_0_devices"] = stage_0_devices
         if limit_mm_per_prompt:
@@ -386,6 +381,7 @@ class AeroRealtimeVLLM(lmms):
                     mm_limits[k.strip()] = int(v)
             else:
                 import json as _json
+
                 mm_limits = {k: int(v) for k, v in _json.loads(limit_mm_per_prompt).items()}
             omni_kwargs["limit_mm_per_prompt"] = mm_limits
 
@@ -394,6 +390,7 @@ class AeroRealtimeVLLM(lmms):
         overrides: dict[str, dict] = {}
         if stage_overrides:
             import json as _json
+
             overrides = _json.loads(stage_overrides)
         if enforce_eager is not None:
             overrides.setdefault("0", {})["enforce_eager"] = bool(enforce_eager)
@@ -466,9 +463,7 @@ class AeroRealtimeVLLM(lmms):
 
     # --------------------------- request -> sample ------------------------------
 
-    def _extract_video_and_text(
-        self, chat: ChatMessages
-    ) -> Tuple[str, str, Optional[float], Optional[float]]:
+    def _extract_video_and_text(self, chat: ChatMessages) -> Tuple[str, str, Optional[float], Optional[float]]:
         """Pull (first_video_path, concatenated_user_text, start_time, end_time) out of ChatMessages."""
         video_path: Optional[str] = None
         start_time: Optional[float] = None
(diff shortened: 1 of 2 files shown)


akawincent:
Note on the second commit (`87eed7cd`, "style: auto-fix lint"): the lint job runs `pre-commit run --all-files` and currently fails on `main` itself — `lmms_eval/models/chat/aero_realtime_vllm.py` (added in #1424) is unformatted, and fork PRs cannot auto-push the fix the way pushes to `main` do. That commit is the exact pre-commit (black + isort) output for that file and touches nothing else. Happy to drop it from this PR if you prefer to format `main` separately.

kcz358 (approved):
Hi @akawincent thanks for the contribution. LGTM! I will merge this PR and thanks for the fix.

PR #1472, opened 2026-08-24
fix(chat-vllm): add is_qwen3_vl flag to route video frames through timestamped encoding

### LINKED ISSUE #1404: [Bug] chat `vllm` model drops video timestamps (to_openai_messages), degrading Qwen3-VL/Qwen3.5 on motion benchmarks
### Summary

The chat `vllm` model (`lmms_eval/models/chat/vllm.py`) encodes video with `ChatMessages.to_openai_messages`, which decodes the clip and emits each frame as a bare `image_url` **with no timestamp / temporal markers**. Qwen3-VL–family models (Qwen3-VL, Qwen3.5) rely on `<t seconds>` timestamp text to know the input is a *video*; without it they treat the frames as a pile of independent images, answer "based on the provided images", and silently underperform on motion/temporal benchmarks.

The repo already contains the correct encoder — `ChatMessages.to_qwen3_vl_openai_messages` — which inserts a `{"type": "text", "text": "<t seconds>"}` marker before each frame. The `vllm` wrapper just never calls it.

This is a *different* code path from #1244 (which fixed the **simple/transformers** backend); the **chat `vllm`** backend has the same class of bug via a different encoder.

### Where

- `lmms_eval/models/chat/vllm.py` → `make_one_request` calls `chat_messages.to_openai_messages(video_kwargs=video_kwargs)`.
- `lmms_eval/protocol.py` → `to_openai_messages` video branch: emits frames as `image_url` with no timestamps.
- `lmms_eval/protocol.py` → `to_qwen3_vl_openai_messages` (already present): the correct timestamped encoding.

### Evidence (Qwen3.5-2B, full benchmarks)

Comparing the native `qwen3_5`/`qwen3_vl` transformers wrapper vs the generic `vllm` wrapper with the default vs the timestamped encoder:

| Benchmark (n) | native (transformers) | vllm default (`to_openai_messages`) | vllm + timestamped encoder |
| --- | --- | --- | --- |
| motionbench_full (4018) | 0.578 | 0.489 | 0.549 |
| mvbench (~4000) | 59.0 | 44.3 | 51.2 |
| mvp_tiny single_acc (4000) | 59.8 | 56.1 | 56.6 (neutral — perception) |

Behaviorally, with the default encoder the model rambles 1000+ chars ("based on the provided images"); with the timestamped encoder it answers tersely like the transformers backend. The fix recovers most of the loss on motion/temporal benchmarks (motionbench +6.0pts, mvbench +6.9pts) and is neutral on static-perception (mvp), i.e. exactly the signature of restored temporal information. (A residual vllm-vs-transformers gap remains, but that is a separate backend-equivalence question.)

### Suggested fix

Route the chat `vllm` wrapper to `to_qwen3_vl_openai_messages` for Qwen3-VL–family video models (e.g. via a `video_timestamps`/model-type flag), instead of the plain `to_openai_messages`.

### Environment

lmms-eval 0.7.1, vllm 0.24.0, transformers 5.14, torchcodec video reader, Qwen3.5-2B. (Please verify against `main`; line numbers may have shifted.)

feiiiiii5:
## Description

Fixes #1404

The chat `vllm` wrapper always encoded video through `ChatMessages.to_openai_messages`, which emits each decoded frame as a bare `image_url` with no temporal markers. Qwen3-VL-family models (Qwen3-VL / Qwen3.5) rely on `<t seconds>` text markers to treat a frame clip as video; without them the frames are read as independent images, and motion/temporal benchmarks silently underperform — exactly as documented in #1404.

This adds an `is_qwen3_vl` boolean to the wrapper (mirroring the existing `async_openai` wrapper's switch, `lmms_eval/models/chat/async_openai.py:63`) and routes both request paths through it:

- `make_one_request` (batched generation)
- `_to_openai_messages` (`generate_until_multi_round`)

Default stays `False`, so non-Qwen3 models are unaffected; users run `--model_args ... is_qwen3_vl=True` for Qwen3-VL-family checkpoints served via vLLM.

Also defers `decord` in `models/simple/vllm.py` via the repo's `optional_import` convention: decord ships no macOS wheels, which made the module unimportable off-Linux (blocking local runs of the wrapper tests). Video-decoding paths still require it at call time.

## Type of Change
- [x] Bug fix

## How I tested

- New `test/models/test_vllm_video_encoder.py`: with both encoders spied, `make_one_request` and `_to_openai_messages` route to `to_qwen3_vl_openai_messages` iff `is_qwen3_vl=True`, else to `to_openai_messages`.
- Existing `test/models/test_vllm_sampling_contract.py` extended for the new attribute: 10/10 pass locally (Python 3.12, vLLM not required — encoder selection is exercised via `VLLM.__new__` following the file's existing pattern).
- black 25.1.0 (--line-length=240, per pre-commit) leaves all touched files unchanged; ruff check clean on changed files.


pufanyi:
Thank you so much for your contribution!

It looks good to me. Seems that linter is failed. I will add a follow up commit to fix the linter and merge the PR.

pufanyi (approved):

PR #1490, opened 2026-08-30
fix: Qwen3-VL pixel budgets to use the 32×32 visual-token grid

gertelrina:
Qwen3-VL uses a 16×16 vision patch size with a 2×2 spatial merger, so one final visual token corresponds to a 32×32 pixel region, not 28×28.

Therefore, the pixel limits in the Qwen3-VL wrapper must use the 32×32 grid. Otherwise, the configured minimum and maximum do not represent the intended number of visual tokens for this model.

This PR updates the grid factor:
it might be (with urs tokens budget)
min_pixels = 256 * 32 * 32
max_pixels = 2048 * 32 * 32

or (with original QWEN's budget)
min_pixels = 64 * 32 * 32
max_pixels = 16384 * 32 * 32

Updating the grid size is required for correct token accounting. The exact maximum visual-token budget may remain at the value the maintainers consider appropriate for evaluation; it only needs to be converted using the correct 32×32 Qwen3-VL grid https://huggingface.co/Qwen/Qwen3-VL-2B-Instruct/blob/main/preprocessor_config.json .

## Summary

- 
- 
- 

## In scope

- 

## Out of scope

- 

## Validation

- 

## Risk / Compatibility

- 

## Type of Change
- [ ] New feature
- [ ] New benchmark/task
- [ ] New model integration
- [ ] Breaking change
- [ ] Documentation update


gertelrina:
@Luodian @kcz358
@pufanyi @jzhang38

Would you prefer to preserve the current minimum and maximum visual-token budgets using the correct 32×32 grid, or should these limits be aligned with the official Qwen3-VL defaults? I would be happy to adjust the PR based on the behavior you consider most appropriate for evaluation?

pufanyi:
Thanks for catching the stale 28×28 factor. I traced the current values back to the Qwen2.5-VL wrapper: `1605632` is `2048 * 28 * 28`, so the existing code represents a nominal 256–2048 visual-token budget.

To keep this PR scoped to correcting the Qwen3-VL grid while preserving the existing lmms-eval budget, could you please update it to:

```python
min_pixels: int = 256 * 32 * 32
max_pixels: int = 2048 * 32 * 32
```

One small note: the lint check is currently failing as well, so please take a look when updating the PR. Thanks!

pufanyi (approved):

kcz358 (approved):

Issue #1505, opened 2026-09-04
mathvision: _fix_sqrt corrupts indexed roots (\sqrt[3]{8} → \sqrt{[}3]{8}), silently failing the latex2sympy equivalence path


feiiiiii5:
## Summary

`_fix_sqrt` in `lmms_eval/tasks/mathvision/eval_utils.py` rewrites braceless `\sqrt` shorthands by bracing the first character after `\sqrt`. For an **indexed root** like `\sqrt[3]{8}`, the first character after `\sqrt` is the `[` of the optional root index — so the rewrite produces `\sqrt{[}3]{8}`, which is invalid LaTeX.

## Failure chain

1. `process_results` → `find_math_answer` → `_strip_string` (eval_utils.py:373) → `_fix_sqrt`.
2. `_fix_sqrt` splits on `\sqrt` and, for each following segment not starting with `{`, braces the first character: `\sqrt[3]{8}` → `\sqrt{[}3]{8}`.
3. Downstream, `is_equal` (eval_utils.py:80) falls back to `latex2sympy` for non-identical strings; the latex2sympy path is wrapped in `except Exception: return False`.

So any gold/model answer containing an indexed root has its LaTeX corrupted before comparison, and the corruption is **converted into a silent false negative** — the exception is swallowed and the sample is scored wrong with no signal.

## Evidence (all run locally on Python 3.11, 2026-09-04)

| step | command | result |
|---|---|---|
| corruption | `_fix_sqrt("\\sqrt[3]{8}")` | returns `"\\sqrt{[}3]{8}"` |
| parseability | `latex2sympy("\\sqrt{[}3]{8}")` | raises `Exception: I expected something else here` (latex2sympy2/latex2sympy2.py:141) |
| clean form parses | `latex2sympy("\\sqrt[3]{8}")` | `8**(1/3)` |
| 异形等价 works uncorrupted | sympy `simplify(8**(1/3) - latex2sympy("8^{1/3}")) == 0` | `True` — so a model answering `8^{1/3}` against gold `\sqrt[3]{8}` would be marked **wrong** even though sympy confirms they are equal |

Pre-fix regression demo (the two failing asserts below pass after the fix):

```
FAILED test_fix_sqrt_preserves_indexed_roots - assert "\\sqrt{[}3]{8}" == "\\sqrt[3]{8}"
FAILED test_strip_string_leaves_indexed_roots_parseable - Exception: I expected something else here
```

## Scope

[...]

PR #1506, opened 2026-09-04
fix(mathvision): preserve optional root indices in _fix_sqrt

feiiiiii5:
Related: #1505

## Summary
- `_fix_sqrt` braced the first character after every `\sqrt` it rewrites; for indexed roots (`\sqrt[3]{8}`) that character is the `[` of the root index, producing invalid LaTeX (`\sqrt{[}3]{8}`)
- The corrupted form makes `latex2sympy` raise inside `is_equal`'s `except: return False` fallback, converting indexed-root comparisons into silent false negatives
- Skip rewriting when the segment already starts with `[`; braceless shorthands (`\sqrt2`, `\sqrta`) keep the existing behavior

## In scope
- `lmms_eval/tasks/mathvision/eval_utils.py`: the guard in `_fix_sqrt` now accepts `{` **or** `[` as "already well-formed" (1-line condition change + comment)
- `test/test_mathvision_eval_utils.py`: regression tests for indexed roots, existing shorthand behavior, and end-to-end `_strip_string` parseability

## Out of scope
- Other task directories (`_fix_sqrt` exists only in mathvision)
- Any change to scoring logic in `is_equal` / `find_math_answer`

## Validation
- `PYTHONPATH=. python -m pytest test/test_mathvision_eval_utils.py -q` | sample size: `N=3` | key metrics: pre-fix 2 failed / 1 passed → post-fix `3 passed` | result: `pass`
- Pre-fix FAIL evidence: `test_fix_sqrt_preserves_indexed_roots` (corruption `\sqrt{[}3]{8}`) and `test_strip_string_leaves_indexed_roots_parseable` (`Exception: I expected something else here` at latex2sympy2.py:141)
- `uvx ruff@0.16.4 check` + `format --check` on both touched files | result: `pass`

## Risk / Compatibility
- No behavior change for `\sqrt{...}` / `\sqrt2` / `\sqrta` inputs; only previously-corrupted indexed-root inputs stop being corrupted
- Verification env: full `lmms-eval` deps installed (Python 3.11), `latex2sympy2` importable

## Type of Change

### DIFF
--- a/lmms_eval/tasks/mathvision/eval_utils.py
+++ b/lmms_eval/tasks/mathvision/eval_utils.py
@@ -308,12 +308,15 @@ def _fix_sqrt(string):
     for split in splits[1:]:
         # If the split portion is non-empty and the first character isn't a '{',
         # then it means the argument of the sqrt is not enclosed in braces.
-        if len(split) > 0 and split[0] != "{":
+        # An optional root index ("\\sqrt[3]{8}") starts with '[' and is already
+        # well-formed: bracing its '[' corrupted it to '\\sqrt{[}3]{8}', invalid
+        # LaTeX that kills the latex2sympy equivalence path in is_equal.
+        if len(split) > 0 and split[0] not in "{[":
             a = split[0]
             # Add braces around the first character and append the rest of the split portion.
             new_substr = "\\sqrt{" + a + "}" + split[1:]
         else:
-            # If the split portion starts with a '{', then it's already correct.
+            # If the split portion starts with a '{' or a root index '[', it's already correct.
             new_substr = "\\sqrt" + split
         # Add the new substring to our result string.
         new_string += new_substr
--- a/test/test_mathvision_eval_utils.py
+++ b/test/test_mathvision_eval_utils.py
@@ -1,0 +1,28 @@
+import pytest
+
+pytest.importorskip("latex2sympy2")
+
+from lmms_eval.tasks.mathvision.eval_utils import _fix_sqrt, _strip_string  # noqa: E402
+
+
+def test_fix_sqrt_preserves_indexed_roots():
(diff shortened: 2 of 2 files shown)


pufanyi:
Thanks for identifying this edge case and for adding regression tests. The helper-level behavior is indeed incorrect when a model response itself contains an indexed root such as `\sqrt[3]{8}`.

I’m hesitant to merge this into `mathvision_standard_eval` at this point, though. This evaluator is a direct port of the [official MATH-V evaluation code](https://github.com/mathllm/MATH-V/blob/main/evaluation/utils.py), which still uses the current behavior. Changing it only in lmms-eval would make our standard metric diverge from the benchmark’s reference implementation and could reduce comparability with the paper and official leaderboard results.

There is also a scope detail in #1505: in the current evaluation path, `find_math_answer` / `_strip_string` is applied only to the model answer, while the gold answer is passed directly to `is_equal`. Therefore, the example with gold `\sqrt[3]{8}` and prediction `8^{1/3}` does not appear to exercise this bug end to end. The current tests demonstrate the helper behavior, but not yet an official sample or released baseline score whose correctness changes.

Could you please first open an issue or PR in the [MATH-V repository](https://github.com/mathllm/MATH-V) and link it here? If the benchmark maintainers confirm or adopt the change, we would be happy to mirror it in lmms-eval. Until then, I would prefer to preserve the official behavior for `mathvision_standard_eval`. If we want to provide a corrected or more permissive evaluator independently, exposing it as a separate versioned metric may be safer.

Thanks again for investigating this carefully.

feiiiiii5:
Filed upstream as mathllm/MATH-V#11 (verified against the reference implementation: `_fix_sqrt('\sqrt[3]{8}')` currently yields `\sqrt{[}3]{8}`). Leaving the reference behavior untouched here until the benchmark maintainers weigh in, per your suggestion.

Issue #1507, opened 2026-09-04
mathvision: _fix_fracs is gated on "sqrt" — \frac12-style shorthand never normalizes, silently failing is_equal for sqrt-free answers


feiiiiii5:
## Summary

In `_strip_string` (`lmms_eval/tasks/mathvision/eval_utils.py`), the call to `_fix_fracs` is guarded by `if "sqrt" in string:` — a copy of the `_fix_sqrt` guard five lines above. Fraction normalization has no interaction with sqrt. The canonical hendrycks `math_equivalence.py` (the reference this chain derives from) calls `_fix_fracs` **unconditionally**.

## Failure chain

1. A MATHVision gold/model answer contains a shorthand fraction (`\frac12`, `\frac1{2}`, `\frac34`) and no `\sqrt`.
2. `_strip_string` skips `_fix_fracs` because of the sqrt guard → the shorthand survives.
3. `is_equal` → `latex2sympy` raises on the shorthand (`Exception: I expected something else here`) → the `except Exception: return False` swallows it → **scored 0**.
4. No exception ever surfaces: a mathematically correct answer in a different surface form (`\frac12` vs gold `\frac{1}{2}`) is a silent false negative.

## Evidence (all run locally, Python 3.11, 2026-09-04, `latex2sympy2` installed)

| input | `_strip_string` output | latex2sympy |
|---|---|---|
| `\frac12` | `\frac12` (unchanged — guard blocks `_fix_fracs`) | **raises** |
| `\frac1{2}` | `\frac1{2}` (unchanged) | **raises** |
| `\frac34` | `\frac34` (unchanged) | **raises** |
| `\frac12 + \sqrt2` | `\frac{1}{2}+\sqrt{2}` (normalized — guard passes) | parses |
| canonical reference | hendrycks `math_equivalence.py` `_strip_string`: `string = _fix_fracs(string)` with no guard | — |

So `_fix_fracs` works whenever it is reachable; the sqrt guard alone blocks it for exactly the strings that need it.

## Scope check

- Ported in with the guard from day one (#527, `70d19c0`).
- `_fix_fracs` in this repo is line-for-line the canonical hendrycks implementation; behavior for `\frac{a}{b}` (braced) strings is unchanged passthrough, so removing the guard only affects unbraced shorthand.
- No open/closed issue or PR covering this (searched `frac in:title,body`, full state).

Aurnawr:
[...]

PR #1508, opened 2026-09-04
fix(mathvision): call _fix_fracs unconditionally in _strip_string

feiiiiii5:
Related: #1507

> Stacked on #1506 (shares its test file and the `_fix_sqrt` hunks; once #1506 merges, the diff of this PR is the frac-guard removal only).

## Summary
- `_strip_string` gated `_fix_fracs` behind `if "sqrt" in string:` — a copy-paste of the `_fix_sqrt` guard; fraction shorthand (`\frac12`, `\frac1{2}`) never normalized for any sqrt-free answer
- The unnormalized shorthand makes `latex2sympy` raise inside `is_equal`'s `except: return False`, silently scoring a correct answer as wrong
- Remove the guard: `_fix_fracs` is called unconditionally, as in the canonical hendrycks `math_equivalence`

## In scope
- `lmms_eval/tasks/mathvision/eval_utils.py`: guard removal (2 deleted lines) + comment
- `test/test_mathvision_eval_utils.py`: four frac tests (normalization without sqrt, end-to-end parseability, canonical passthrough for `\frac{1}2`, control with sqrt present)

## Out of scope
- The `\frac{1}2` form (braced numerator): canonical `_fix_fracs` leaves it as-is; kept as a parity test, not changed
- Any scoring logic in `is_equal` / `find_math_answer`

## Validation
- `PYTHONPATH=. python -m pytest test/test_mathvision_eval_utils.py -q` | sample size: `N=7` | key metrics: pre-fix 2 failed / 5 passed → post-fix `7 passed` | result: `pass`
- Pre-fix FAIL evidence: `test_frac_shorthand_normalized_without_sqrt` (`\frac12` survived `_strip_string`) and `test_frac_shorthand_parseable_without_sqrt` (`Exception: I expected something else here` from latex2sympy2.py:141)
- Control: `\frac12 + \sqrt2` normalized correctly both before and after (proves the guard was the only blocker)
- `uvx ruff@0.16.4 check` + `format --check` on both touched files | result: `pass`

## Risk / Compatibility
- Braced `\frac{a}{b}` strings pass through `_fix_fracs` unchanged → zero behavior delta for already-well-formed answers
- Strings containing `sqrt` behave identically to before (the function already ran there)
- Only unbraced shorthand changes: exactly the inputs that previously caused silent false negatives

## Type of Change

### DIFF
--- a/lmms_eval/tasks/mathvision/eval_utils.py
+++ b/lmms_eval/tasks/mathvision/eval_utils.py
@@ -308,12 +308,15 @@ def _fix_sqrt(string):
     for split in splits[1:]:
         # If the split portion is non-empty and the first character isn't a '{',
         # then it means the argument of the sqrt is not enclosed in braces.
-        if len(split) > 0 and split[0] != "{":
+        # An optional root index ("\\sqrt[3]{8}") starts with '[' and is already
+        # well-formed: bracing its '[' corrupted it to '\\sqrt{[}3]{8}', invalid
+        # LaTeX that kills the latex2sympy equivalence path in is_equal.
+        if len(split) > 0 and split[0] not in "{[":
             a = split[0]
             # Add braces around the first character and append the rest of the split portion.
             new_substr = "\\sqrt{" + a + "}" + split[1:]
         else:
-            # If the split portion starts with a '{', then it's already correct.
+            # If the split portion starts with a '{' or a root index '[', it's already correct.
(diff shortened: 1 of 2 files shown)


feiiiiii5:
Ping on the companion fix — checks green, no conflicts. Calling _fix_fracs unconditionally closes the path where a stranded \\frac survived stripping. Fine to take these two in either order.

Issue #1509, opened 2026-09-04
mathvision: find_math_answer deletes the \text command name but keeps its braces — \boxed{\text{Sunday}} is scored wrong against gold Sunday


feiiiiii5:
## Summary

`find_math_answer` (`lmms_eval/tasks/mathvision/eval_utils.py:420`) cleans the extracted answer with `ans.replace("\\text", "")` — it deletes the `\text` command name but **leaves the braces behind**: `\boxed{\text{Sunday}}` becomes `{sunday}`. Braced non-numeric content cannot be post-processed by `latex2sympy`, so `is_equal`'s `except: return False` turns every text answer (dates, names, unit-bearing values) into a **silent false negative** against a bare gold string.

## Evidence (Python 3.11, 2026-09-04, `latex2sympy2` installed)

| step | input | output |
|---|---|---|
| extraction | `find_math_answer("\\boxed{\\text{Sunday}}")` | `{sunday}` (braces left behind) |
| scoring | `is_equal("{sunday}", "Sunday")` | **False** — string paths don't match, `latex2sympy("{sunday}")` raises, `except → False` |
| lm-eval precedent | proper unwrap `re.sub(r"(\\\\text\\{)(.*?)(\\})", "\\\\2", ...)` | `Sunday` — string-equal to gold → True |
| already-working case | `is_equal("{8}", "8")` | True (numeric single-brace parses) — must not regress |
| nested case | `latex2sympy("{\\\\frac{1}{2}}")` | parses → 1/2 — the legacy bare-replace result for nested `\text` must be preserved |

So the fix must unwrap brace-balanced `\text{...}`/`\mbox{...}` spans while leaving nested ones on the legacy path.

## Scope

- Pattern is unique to `mathvision/eval_utils.py` (no other task file does `replace("\\text", "")`).
- No open/closed issue or PR covering this (searched full state; #660 is an unrelated closed 2024 PR).

Aurnawr:
Hi! I'd like to fix this along with #1505/#1507/#1511 in a single PR, since they all live in mathvision/eval_utils.py. Will open it shortly.

PR #1510, opened 2026-09-04
fix(mathvision): unwrap \\text{...}/\\mbox{...} instead of leaving braces behind

feiiiiii5:
Related: #1509

> Stacked on #1506 → #1508 (shares their test file; each earlier PR is green and merges independently — once they merge, this diff shrinks to the unwrap change).

## Summary
- `find_math_answer` deleted the `\text` command name but kept its braces: `\boxed{\text{Sunday}}` → `{sunday}`, which `latex2sympy` cannot post-process, so `is_equal` scored False against the bare gold `Sunday`
- Every braced text answer (dates, names, unit-bearing values) was a silent false negative
- Unwrap brace-balanced `\text{...}`/`\mbox{...}` spans; nested spans fall back to the legacy bare replace (`{\frac{1}{2}}` parses and must keep doing so)

## In scope
- `lmms_eval/tasks/mathvision/eval_utils.py`: one line becomes a regex unwrap + legacy fallback + comment
- `test/test_mathvision_eval_utils.py`: four tests (text answer unwrap + end-to-end is_equal, mbox unwrap, nested fallback parity, plain-boxed control)

## Out of scope
- The greedy boxed-extraction regex (`oxed{(.*)}`)
- Other task directories (pattern is unique to mathvision)
- `\textbf{A}` handling (unchanged legacy behavior, outcome identical)

## Validation
- `PYTHONPATH=. python -m pytest test/test_mathvision_eval_utils.py -q` | sample size: `N=11` | key metrics: pre-fix 2 failed / 9 passed → post-fix `11 passed` | result: `pass`
- Pre-fix FAIL evidence: `test_boxed_text_answer_is_unwrapped_not_braced` (`{sunday}` vs expected `sunday`; `is_equal` False) and `test_mbox_answer_is_unwrapped` (`{8}` vs expected `8`)
- Parity controls pass before AND after: nested `\text{\frac{1}{2}}` keeps the legacy `{\frac{1}{2}}` result (parses), plain `\boxed{3}` untouched
- `uvx ruff@0.16.4 check` + `format --check` | result: `pass`

## Risk / Compatibility
- Non-nested `\text{...}`/`\mbox{...}`: braces removed — strictly closer to gold for text answers; numeric forms like `{8}` compared equal either way
- Nested forms: byte-identical to before (fallback path)
- No change to `is_equal` / extraction regex / other tasks

## Type of Change

### DIFF
--- a/lmms_eval/tasks/mathvision/eval_utils.py
+++ b/lmms_eval/tasks/mathvision/eval_utils.py
@@ -417,7 +417,16 @@ def find_math_answer(s: str) -> str:
     # Clean the string from various LaTeX formatting.
     ans = ans.replace(" ", "").replace("\\,", "").replace("∞", "\\infty")
     ans = ans.replace("+\infty", "\infty").replace("\\\\", "\\").replace("\n", "")
-    ans = ans.replace("\\text", "").replace("\\mbox", "").replace("bmatrix", "pmatrix")
+    # Unwrap \text{...}/\mbox{...} content instead of deleting only the
+    # command name: leftover braces corrupt the answer (\text{Sunday} ->
+    # "{sunday}", which latex2sympy cannot post-process, so is_equal
+    # scores False against the bare gold "Sunday"). Only brace-balanced
+    # spans are unwrapped; anything with nested braces falls back to the
+    # legacy bare replace, because "{\frac{1}{2}}" parses and must keep
+    # doing so.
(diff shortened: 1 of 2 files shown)


feiiiiii5:
Bumping this — still mergeable, checks green. The fix unwraps \\text/\\mbox wrappers instead of leaving braces behind in the stripped answer. Let me know if you'd rather handle these cases differently.

Issue #1511, opened 2026-09-04
mathvision: find_math_answer keeps the FIRST \\boxed answer when a model retracts it and boxes a corrected one


feiiiiii5:
## Summary

`find_math_answer` extracts the boxed answer with a **greedy** regex `oxed{(.*)}` (`re.S`): the match spans from the first `\boxed{` to the last `}` of the whole string. With a self-correcting model that emits a box, retracts it, and boxes the final answer, the `}-split` fallback below resurrects the **first, retracted** answer — while the code's own `findall(s)[-1]` indexing (and lm-eval's `last_boxed_only_string` convention) intend the **last**.

## Evidence (Python 3.11, 2026-09-04)

```
>>> find_math_answer("the answer is \\boxed{2}. wait, actually \\boxed{3}")
'2'          # the retracted intermediate; intended: '3'
>>> find_math_answer("\\boxed{\\frac{1}{2}}")
'\\frac{1}{2}'   # nested single box is currently correct — must not regress
```

Scoring consequence: a reasoning model that says `\boxed{2}` then corrects itself to `\boxed{3}` is scored against `2` → silent false negative against gold `3`.

## Scope

- Unique to `lmms_eval/tasks/mathvision/eval_utils.py`; no other task uses this extraction.
- No open/closed issue or PR covering multi-box extraction (searched `boxed in:title,body`, full state).

Aurnawr:
Hi! I'd like to fix this along with #1505/#1507/#1509  in a single PR, since they all live in mathvision/eval_utils.py. Will open it shortly.

PR #1512, opened 2026-09-04
fix(mathvision): extract the last \\boxed{...} with balanced braces

feiiiiii5:
Related: #1511

> Stacked on #1506 → #1508 → #1510 (shares their test file; each earlier PR is green and merges independently — this diff shrinks as they merge).

## Summary
- The greedy `oxed{(.*)}` regex spans from the first `\boxed{` to the last `}` of the string; with a retracted intermediate box, the `}-split` fallback resurrected the FIRST answer while the code's own `[-1]` indexing (and lm-eval's `last_boxed_only_string` convention) intend the last
- Replace with a depth-balanced extraction of the last `\boxed{...}` span; no balanced box → whole-string fallback (unchanged)

## In scope
- `lmms_eval/tasks/mathvision/eval_utils.py`: new `_last_boxed_content` helper + call site swap
- `test/test_mathvision_eval_utils.py`: four tests (last-box wins, nested single-box parity, no-box fallback, plain box control)

## Out of scope
- `\fbox` support (the old regex never matched it either)
- Other task directories

## Validation
- `PYTHONPATH=. python -m pytest test/test_mathvision_eval_utils.py -q` | sample size: `N=15` | key metrics: pre-fix 1 failed / 14 passed → post-fix `15 passed` | result: `pass`
- Pre-fix FAIL evidence: `test_last_boxed_wins_over_retracted_intermediate` (`'2'` != `'3'` — retracted intermediate won)
- Parity controls pass before AND after: nested `\boxed{\frac{1}{2}}` content preserved (the greedy regex was right there), `no box here` → `noboxhere` fallback, plain `\boxed{7}`
- `uvx ruff@0.16.4 check` + `format --check` | result: `pass`

## Risk / Compatibility
- Single-box responses: byte-identical output (balanced extraction matches the greedy span for one box)
- Multi-box responses: now the LAST box — strictly the convention the `[-1]` index and lm-eval already document
- No change to `is_equal` / `_strip_string`

## Type of Change

### DIFF
--- a/lmms_eval/tasks/mathvision/eval_utils.py
+++ b/lmms_eval/tasks/mathvision/eval_utils.py
@@ -417,7 +417,16 @@ def find_math_answer(s: str) -> str:
     # Clean the string from various LaTeX formatting.
     ans = ans.replace(" ", "").replace("\\,", "").replace("∞", "\\infty")
     ans = ans.replace("+\infty", "\infty").replace("\\\\", "\\").replace("\n", "")
-    ans = ans.replace("\\text", "").replace("\\mbox", "").replace("bmatrix", "pmatrix")
+    # Unwrap \text{...}/\mbox{...} content instead of deleting only the
+    # command name: leftover braces corrupt the answer (\text{Sunday} ->
+    # "{sunday}", which latex2sympy cannot post-process, so is_equal
+    # scores False against the bare gold "Sunday"). Only brace-balanced
+    # spans are unwrapped; anything with nested braces falls back to the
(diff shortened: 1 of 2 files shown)


feiiiiii5:
Friendly nudge: this mathvision scoring series (#1506, #1508, #1510, #1512 — balanced-brace boxed extraction, text/mbox unwrap, unconditional frac fix, root-index preservation) is ready on my side with regression tests and green CI. If a single squashed PR would be easier to review than four, say the word and I'll consolidate.

PR #1513, opened 2026-09-05
fix(cache): validate recovered responses and repair invalid entries

Luodian:
We found that reopening the response cache could restore empty answers that the live write path had rejected. This applies the same response checks during recovery, reads, and shard merging, so failed responses cannot become cache hits and valid retries can repair old entries.

## Root cause

The audit log records every response before SQLite validation. Replay trusted the deterministic flag and cache key without validating the stored response. Database reads also trusted existing payloads, and shard merging preserved invalid destination entries.

## Fix

- Validate persisted JSON with the existing response validity policy during replay and local/shared reads.
- Keep invalid responses in the audit log without a reusable cache key. Continue validating legacy keyed records.
- Recover valid audit responses over invalid rows and merge valid retries over invalid shared entries. Preserve existing valid responses and skip invalid shard inputs.

## Minimal reproduction

The six added TestCrashRecovery methods run against the base response-cache module at e0bfc699 with nine assertion failures, including subtests. They exercise empty/null answers, malformed likelihood responses, invalid stored JSON, valid audit recovery, and shared-cache repair.

## Validation

- `uv run python -m pytest test/cache -q`: 61 passed, 4 subtests passed, using the existing local environment.
- The shared-cache regression verifies miss -> retry -> merge -> fresh-run hit without another model call.
- `uvx pre-commit run --all-files`: passed.

## Risk / Compatibility

Cache keys and the deterministic-request policy are unchanged. This uses the existing validity policy, including its two-element shape check for likelihood responses; it does not add element-type validation. Tests use mocked models and temporary SQLite/JSONL files.

### DIFF
--- a/docs/advanced/caching.md
+++ b/docs/advanced/caching.md
@@ -101,7 +101,7 @@ To force re-evaluation: delete the `{model_hash}/` directory under your cache pa
 
 ### Crash recovery
 
-Write order: JSONL append + fsync -> SQLite upsert. On startup, any JSONL entries missing from SQLite are replayed. This survives crashes between the two writes.
+Write order: JSONL append + fsync -> SQLite upsert. On startup, valid JSONL entries missing from SQLite are replayed. Valid audit entries also replace invalid SQLite entries left by older versions. This survives crashes between the two writes.
 
 ### Poisoning prevention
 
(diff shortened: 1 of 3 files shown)


pufanyi (approved):

pufanyi:
GPT suggested fixing the legacy audit-log merge case where an invalid entry can cause a later valid retry to be dropped. Personally, I’m fine leaving this edge case unfixed and don’t consider it a blocker for this PR.

PR #1516, opened 2026-09-07
feat: add Atlas Cloud LLM judge provider

binyangzhu000-sudo:
## Summary
- Add Atlas Cloud as an explicit OpenAI-compatible LLM judge provider.
- Keep existing OpenAI and default provider behavior unchanged.
- Document configuration and cover registration/environment isolation with tests.

## In scope
- AtlasCloudProvider, factory registration, environment configuration, documentation, and focused unit tests.

## Out of scope
- Async Atlas Cloud support, task/model backends, and changes to retry policy.

## Validation
- uv run --no-project --with pytest --with loguru --with pillow --with requests --with openai --with aiohttp python -m pytest -q --confcutdir=test/llm_judge test/llm_judge/test_atlascloud_provider.py | sample size: N=3 | result: pass
- uvx ruff check plus uvx ruff format --check on all changed Python files | sample size: N=5 files | result: pass
- Live Atlas Cloud chat smoke with openai/gpt-4.1-mini | sample size: N=1 | expected exact response returned

## Risk / Compatibility
- Low risk: Atlas Cloud is opt-in and OpenAI environment behavior is regression-tested.

## Type of Change
- [x] New feature
- [ ] New benchmark/task
- [ ] New model integration
- [ ] Breaking change
- [x] Documentation update


pufanyi:
Thanks for the contribution and for taking the time to add tests and documentation.

Atlas Cloud exposes an OpenAI-compatible endpoint, so it can already be used through the existing `openai` provider:

```bash
export API_TYPE=openai
export OPENAI_API_KEY="<ATLASCLOUD_API_KEY>"
export OPENAI_API_URL="https://api.atlascloud.ai/v1"
```

Users can then configure an Atlas Cloud model ID such as `openai/gpt-4.1-mini`.

Since the existing provider already supports this use case and the proposed class does not add provider-specific protocol behavior, we prefer to avoid adding a dedicated class and registry entry for each OpenAI-compatible service. We’ll close this PR, but we appreciate the contribution and the careful validation work.

PR #1480, opened 2026-08-28
fix(vllm): avoid eager Decord loading

pufanyi:
## Summary

- Prevent Decord from being imported when either the default chat vLLM route or the `--force_simple` vLLM route is merely resolved/imported, removing the native-library collision precondition reported in #1432 for image-only evaluations.
- Route simple-vLLM video decoding through the shared loader, defaulting to PyAV while retaining Decord as an explicit lazy legacy backend for reproducibility.
- Make PyAV fallback sampling correct for containers whose metadata reports `stream.frames == 0`, and preserve the historical vLLM short-video frame-count behavior at the adapter boundary.
- Add fresh-process import guards, real-container decoder tests, and short-video compatibility coverage.

Fixes #1432.

## Root Cause

The first revision removed Decord from `lmms_eval.models.simple.vllm`, but `--model vllm` resolves to the chat wrapper by default. That route imports `protocol.py`, which still eagerly imported two unused Decord symbols. As a result, the initial fix protected only `--force_simple`; the default route still mapped `libdecord.so`.

Decord 0.6.0 loads its native library globally and exports `__cxa_thread_atexit`. The reported teardown crash can therefore be triggered even by an image-only evaluation that never needs a video decoder.

### Import and backend isolation

- Remove the unused Decord imports from `protocol.py`.
- Lazily import PyAV, Decord, TorchCodec, or DALI only when the selected backend is actually used.
- Add `decord` to the shared decoder dispatcher as an explicit legacy backend.
- Keep `load_video_decord` and its duplicate-index behavior for existing adapters.

### Sampling correctness

- Use one uniform NumPy-linspace index policy across the random-access PyAV, TorchCodec, and Decord paths.
- When PyAV cannot trust the container frame count, decode packets to determine the real count and then sample the requested number of frames instead of returning the whole video.
- Reopen the PyAV container before packet fallback so a failed stream decode cannot leave a reused codec context in a partial state.
- Return unique source frames from the shared loader for short videos, then repeat them inside the simple-vLLM adapter to reproduce the historical Decord output-count contract without changing every shared-loader caller.

## Reproduction / Compatibility

Decord is **not removed**. Historical runs can pin it explicitly:

- Simple vLLM/shared loader: `--force_simple --model_args ...,video_decode_backend=decord`, or `LMMS_VIDEO_DECODE_BACKEND=decord`, with the `video-legacy` extra.
[...]

PR #1517, opened 2026-09-08
fix: change decord to lazy import

microslaw:
## Summary

- Decord is an old library (no activity for the past 5 years)
- When running `import decord` libdecord is loaded, which overrides some references causing segfaults on teardown

## In scope

- Moves all decord imports from top of the files to functions so that it is not loaded unless absolutely needed. This breaks PEP-8 guide for imports, but the precedents for this were already present within the repository and pr #1480 introduced more of those 

## Out of scope

- This pr does not fully remove decord as a backend, or modify any functionality

## Validation

- Ran full test suite

## Risk / Compatibility

- No breaking changes, delayed imports should not break anything

## Type of Change
- [ ] New feature
- [ ] New benchmark/task
- [ ] New model integration
- [ ] Breaking change
- [ ] Documentation update


pufanyi (approved):
Hello! I'm ok with it. I will merge it. Thanks for your contribution!

pufanyi:
Seems that there is a linter error, maybe I can help to fix it.

PR #1520, opened 2026-09-13
fix: bound memory during PyAV packet sampling

### LINKED ISSUE #1519: PyAV packet fallback retains the entire decoded video before frame sampling
### Summary

`read_video(..., num_frm=8, backend="pyav")` returns only eight frames, but the WebM/Matroska packet path first retains every decoded `VideoFrame` in `record_video_length_packet`. The same happens when the stream path falls back after a decode/frame-count failure. Memory therefore scales with video duration even when the model's input frame budget is small.

### Reproduction

Create a 300-frame, 640×360 Matroska video with PyAV's FFV1 encoder, then run `read_video(path, num_frm=8)` in a fresh Python process. The generated file was 181,809 bytes, while peak process RSS was **186.6 MiB** on macOS with PyAV 18.1.0. The growth occurs before RGB conversion: `load_video_packet` builds the full decoded frame list before applying uniform indices.

A deterministic regression test with 200 generated frame objects also observes all 200 alive at once while requesting just four. This is reachable through the shared loader used by video adapters such as LLaVA-OneVision, and is especially costly for long-video evaluation.

### Proposed fix and validation

Count decoded frames without keeping them, rewind the container, and collect only the existing uniform sample on a second packet pass. This preserves frame indices, FPS limits, short-video behavior and the final-frame option, at the cost of a second decode pass on the fallback path.

The local fix reduced the same synthetic run to **79.0 MiB** peak RSS with the same eight frame values. Twenty-two loader tests pass, including real WebM, Matroska and MP4 fallback decoding, FPS/last-frame cases and a live-frame retention bound. No model weights or benchmark datasets are needed to reproduce this.

pentaoa:
## Summary
- Fixes #1519: PyAV's packet fallback retained the entire decoded video before selecting a small frame sample, so long-video evaluation could exhaust memory despite a low `num_frm` setting.
- Count frames without retaining them, rewind, then collect the existing uniform sample. This covers WebM/Matroska and fallback after a failed stream decode.
- A 300-frame 640×360 synthetic MKV (181,809 bytes on disk), sampled to eight frames in separate processes, used **186.6 MiB → 79.0 MiB** peak RSS with identical sampled frame values. This is one small CPU measurement, not a model benchmark.

## In scope
- Replace the list-building packet scan with an iterator and two-pass sampling.
- Preserve uniform indices, FPS caps, short-video behavior, decoder flush frames and `force_include_last_frame`.
- Add live-frame retention regressions and real WebM/MKV/MP4 fallback sampling tests.

## Out of scope
- Other decode backends, model adapters and benchmark scoring.

## Validation
- `python -m pytest test/models/test_load_video.py -q` | sample size: `N=22` tests | key metrics: 22 passed, including real video decoding with PyAV 18.1.0; three retention cases fail before the fix | result: pass
- `pre-commit run --files lmms_eval/models/model_utils/load_video.py test/models/test_load_video.py` | sample size: `N=2` files | key metrics: Ruff check and format pass | result: pass
- `pre-commit run --all-files` | sample size: repository | key metrics: initially required an unrelated import reorder in `test/eval/test_physics_benchmarks.py`; that automatic change was reverted and is not included here | result: baseline lint issue

## Risk / Compatibility
- The packet fallback now decodes twice and requires rewinding its local input. This deliberately trades decode work for memory bounded by the requested sample plus decoder buffering. The successful metadata-based stream path remains a single pass.
- No model weights or evaluation datasets were downloaded. The existing sampling policy and returned array shape/dtype are preserved.

## Type of Change
- [ ] New feature
- [ ] New benchmark/task
- [ ] New model integration
- [ ] Breaking change
- [ ] Documentation update


pufanyi:
Thanks for the contribution! Overall, the two-pass decoding approach looks good and addresses the unbounded memory usage while preserving the existing sampling policy.

I found one blocker in the current repository environment: the project is pinned to `av<16.0.0`, with PyAV 15.1.0 in `uv.lock`, while this PR was validated with PyAV 18.1.0. Running the new test suite against the locked environment gives 3 failures in the MP4/MPEG-4 cases (`19 passed, 3 failed`), because the tests assume the decoded pixel means remain within an absolute tolerance of 2; with the bundled codec version, the expected value of 110 decodes to approximately 107.33.

We’re planning to upgrade the project’s PyAV version soon, so I’d prefer to revisit this PR once that dependency upgrade lands and rerun it against the new locked environment. Alternatively, if we want to merge this beforehand, the MP4 assertions should be made codec/version-independent—for example, by comparing the sampled frames against a reference decode of the same encoded video.

One small non-blocking suggestion: after the second decode pass, it would also be helpful to verify that `len(frames) == len(indices)`, so an incomplete rewind or decode cannot silently return fewer frames.

Aside from that, the implementation and memory-retention tests look good. Thanks again!

pufanyi:
Quick update: I opened #1536 to upgrade PyAV to 17.1, and its CI is passing. I also reran this PR's expanded video tests with PyAV 17.1.0; the result is still 3 failed / 19 passed.

All three failures are the MP4/mpeg4 cases: the decoded last-frame mean is 107.33, while the test expects 110 with . This looks like normal lossy codec/FFmpeg variation rather than an issue with the sampling logic.

Could you make those assertions a little more robust across FFmpeg builds—for example, by using a codec-specific tolerance of around 3–4? Once that is addressed, this should be ready to approve.

pufanyi:
Maybe I can try to fix it

pufanyi:
It looks like Astra 大人 edited the test file directly, but I took a look and there doesn't seem to be any major issue. Maybe I will merge it.

pufanyi (approved):

PR #1529, opened 2026-09-15
fix(llava): handle transformers compat for apply_chunking_to_forward import

### LINKED ISSUE #860: Error while runnig lava-onevision-qwen2-7b example
```
export HF_HOME="~/.cache/huggingface"

accelerate launch --num_processes=8 --main_process_port 12399 -m lmms_eval \
    --model=llava_onevision \
    --model_args=pretrained=lmms-lab/llava-onevision-qwen2-7b-ov,conv_template=qwen_1_5,device_map=auto,model_name=llava_qwen \
    --tasks=longvideobench_val_v \
    --batch_size=1 \
    --verbosity=DEBUG
```

When I try to run the code above, I got the error above:

```
Failed to import llava_llama from llava.language_model.llava_llama. Error: cannot import name 'apply_chunking_to_forward' from 'transformers.modeling_utils' (~/anaconda3/envs/lmms-eval/lib/python3.12/site-packages/transformers/modeling_utils.py)
Failed to import llava_qwen from llava.language_model.llava_qwen. Error: cannot import name 'apply_chunking_to_forward' from 'transformers.modeling_utils' (~/anaconda3/envs/lmms-eval/lib/python3.12/site-packages/transformers/modeling_utils.py)
Failed to import llava_mistral from llava.language_model.llava_mistral. Error: cannot import name 'apply_chunking_to_forward' from 'transformers.modeling_utils' (~/anaconda3/envs/lmms-eval/lib/python3.12/site-packages/transformers/modeling_utils.py)
Failed to import llava_mixtral from llava.language_model.llava_mixtral. Error: cannot import name 'apply_chunking_to_forward' from 'transformers.modeling_utils' (~/anaconda3/envs/lmms-eval/lib/python3.12/site-packages/transformers/modeling_utils.py)
```

lorenzozanee:
## Summary
- Fix `llava_onevision` (and sibling llava models) failing to import on transformers>=4.47 where `apply_chunking_to_forward` moved from `modeling_utils` to `pytorch_utils`
- Add `lmms_eval/models/model_utils/llava_transformers_compat.py` that mirrors `apply_chunking_to_forward`, `find_pruneable_heads_and_indices`, `prune_linear_layer` bidirectionally and shims `PreTrainedModel` for mistaken imports
- Call the shim before any `from llava.* import` in `llava_onevision`, `llava`, `llava_vid`, `vila`, `slime`, `llava_onevision_moviechat`

## In scope
- New compat module + 6 one-line ensures in llava-family model wrappers

## Out of scope
- No change to model logic, training, or evaluation; no dependency pin
- Does not vendor or patch external LLaVA code itself

## Validation
- `python /tmp/test_compat.py` with mocked new/old transformers layout | N=2 scenarios | result: pass
- `ruff check` on patched files | result: pass
- `python -m py_compile` on patched files | result: pass

## Risk / Compatibility
- No breaking change; shim is no-op when both locations already have helpers
- Isolated to llava-family imports; does not affect `llava_hf` path

## Type of Change

Fixes #860

Supercedes #1497（原PR分支已删，按原提交重建）


pufanyi:
Thanks for working on this. I think the underlying issue is valid, but this PR needs a few changes before it is ready to merge:

1. The shim does not cover the currently allowed Transformers 5.x versions. `transformers` is only constrained to `>=4.39.2`, while `find_pruneable_heads_and_indices` has been removed entirely in Transformers 5.x. Since neither module contains it, the original LLaVA-NeXT import will still fail in a fresh environment.
2. `llava_vid.py` reintroduces a module-level `decord` import. The existing `load_video.py` explicitly avoids this because it can cause an exit-time SIGSEGV on some accelerator environments. This import is also unused and should be removed.
3. I do not think we should inject `PreTrainedModel` into `pytorch_utils`. As noted in [the issue discussion](https://github.com/EvolvingLMMs-Lab/lmms-eval/issues/860#issuecomment-3924091938), that error comes from an incorrect downstream edit and is separate from the three moved helper functions. The compatibility patch should remain narrowly scoped to the original issue.
4. The current lint check is failing, and the validation only references an uncommitted `/tmp/test_compat.py`. Could you please add a reproducible regression test and run a real end-to-end evaluation with an actual LLaVA-NeXT installation and Transformers 4.57.x, ideally with a small task or `--limit 1`? Please also document the expected behavior with Transformers 5.x.

A narrow one-way compatibility alias for the three helpers may be reasonable for Transformers 4.56–4.57, but we should avoid claiming compatibility regardless of the installed Transformers version until the default 5.x environment is handled or explicitly rejected. Thanks!

kcz358:
Hi, thank you for the contribution. I think I agree for the suggestions from fanyi and if you are using a newer transformers version for evaluating llava, I think using llava_hf would be a more suitable model type. Thanks!

